PR #32, opened 2026-03-31
feat: add canvas list and read commands

### LINKED ISSUE #31: feat: add canvas list and read commands
With AI agents increasingly relying on CLI tools over MCP — especially when MCP requires workspace admin approval — slackcli hits a sweet spot for Slack automation. I've been using it to build agents that monitor conversations and it works great.

The one gap I'm hitting is Canvas. My agents can read messages but can't access documentation that teams put in Canvas. Since slackcli already positions itself as AI-friendly, canvas support feels like a natural next step.

### Proposal

Add a `canvas` command group:

```bash
slackcli canvas list [--limit] [--channel] [--json]
slackcli canvas read <canvas-id> [--channel] [--raw] [--json]
```

The Slack API has no `canvases.read` method, but canvas content can be retrieved as HTML via `files.info` + download — this is the established workaround used by [agent-slack](https://github.com/stablyai/agent-slack), [slack-explorer-mcp](https://github.com/shibayu36/slack-explorer-mcp), and others.

The HTML-to-Markdown conversion would be an internal module with zero new dependencies, following the same pattern as `mrkdwn.ts`.

### Context

I'm aware of #7 which was closed for inactivity. My approach differs: zero deps, focused scope (no channel name resolution), and Slack-specific HTML handling (checklists, emoji, mentions, embedded files).

I have a working implementation on my fork — happy to open a PR if this is something you'd welcome.

jiel:
## Summary

- Add `canvas list` and `canvas read` commands to list and read Slack Canvas documents as Markdown
- Zero new dependencies — HTML-to-Markdown conversion via internal `canvas-parser.ts` module
- Dual auth support (standard + browser tokens), consistent with the rest of the codebase

Closes #31

## Commands

```bash
# List canvases
slackcli canvas list
slackcli canvas list --limit=50 --channel=C1234567890 --json

# Read canvas content
slackcli canvas read F1234567890
slackcli canvas read F1234567890 --json    # includes markdown, edit_timestamp, editors
slackcli canvas read F1234567890 --raw     # raw HTML, no conversion
slackcli canvas read --channel=C1234567890 # read channel canvas
slackcli canvas read --channel=D09B6A9R91Q # also works for DMs
```

## How it works

The Slack API has no `canvases.read` method. Canvas content is retrieved via `files.info` → `url_private_download` → HTML download, then converted to Markdown. This is the same workaround used by [agent-slack](https://github.com/stablyai/agent-slack) and [slack-explorer-mcp](https://github.com/shibayu36/slack-explorer-mcp).

The HTML is Quip-based with Slack-specific structures. The parser handles:

| Feature | HTML source | Markdown output |
|---------|------------|-----------------|
| Checklists | `data-section-style='7'` + `class="checked"` | `- [x]` / `- [ ]` |
| Ordered lists | `data-section-style='6'` (uses `<ul>` not `<ol>`) | `1.` numbered |
| Custom emoji | `<control data-remapped>` | `:emoji_name:` |
| @mentions | `<control><a>@U...</a></control>` | Resolved to `@Display Name` |
| #channels | `<control><a>#C...</a></control>` | Resolved to `#channel-name` |
| Embedded files | `<p class="embedded-file">` | `[filename](url)` |
| Code blocks | `<pre class="prettyprint">` with internal `<i>`/`<b>` | Fenced blocks (tags stripped) |
| Tables | `<table>` with `<h1>`/`<b>` in cells | GFM tables with formatting preserved |

The `--json` output includes `edit_timestamp` (last content edit) and `editors` (user IDs who have edited), useful for tracking whether a canvas has been updated.

The `--channel` option works for public channels, private channels, and DMs (canvas lookup checks `properties.canvas`, `properties.meeting_notes`, and `properties.tabs`).

Guards: 10 MB file size limit, auth page detection (expired token), body stream cancel on oversized responses.

### New files
| File | Lines | Purpose |
|------|-------|---------|
| `src/lib/canvas-parser.ts` | 377 | HTML→Markdown converter (zero deps, 10-step pipeline) |
| `src/lib/canvas-parser.test.ts` | 342 | 37 tests with inline HTML fixtures from real Canvas exports |
| `src/commands/canvas.ts` | 227 | CLI commands (list + read) with mention resolution |

### Modified files
| File | Change |
|------|--------|
| `src/lib/slack-client.ts` | +4 methods: `listCanvases`, `getFileInfo`, `downloadFile`, `getChannelCanvasId` |
| `src/types/index.ts` | +3 interfaces: `SlackCanvas`, `CanvasListOptions`, `CanvasReadOptions` |
| `src/lib/formatter.ts` | +2 functions: `formatCanvasList`, `formatCanvasContent` |
| `src/index.ts` | Register `createCanvasCommand` |
| `README.md` | Canvas Commands section + feature badge |
| `CLAUDE.md` | Updated command count, added `canvas-parser.ts` to module table |

## Test plan

- [x] `bun test` — 109 tests pass (72 existing + 37 new), 0 fail
- [x] `bun run type-check` — 0 errors
- [x] Manual e2e: `slackcli canvas list` on real workspace
- [x] Manual e2e: `slackcli canvas read F<id>` on real workspace
- [x] Manual e2e: `slackcli canvas read F<id> --json`
- [x] Manual e2e: `slackcli canvas read --channel C<id>` (public channel)
- [x] Manual e2e: `slackcli canvas read --channel D<id>` (DM)

## Relation to #7

This builds on the groundwork laid by #7, which identified the `files.info` download approach and demonstrated the demand for canvas support. I've incorporated the feedback from that review (zero new deps, input validation, size guard, focused scope) and added Slack-specific HTML handling for checklists, emoji, and mentions. Happy to credit @brianevanmiller for the prior work.


src/lib/slack-client.ts
```diff
@@ -301,6 +301,69 @@ export class SlackClient {
...
+
+    // Size guard
+    const contentLength = response.headers.get('content-length');
+    if (contentLength && parseInt(contentLength, 10) > maxBytes) {
+      await response.body?.cancel();
+      throw new Error(`File too large: ${contentLength} bytes (max ${maxBytes})`);
+    }
+
+    return response.text();
+  }
```
shahariaazam, line 373: **🟡 MUST FIX: Size guard is bypassable**

`Content-Length` is an optional HTTP header — chunked transfer encoding omits it entirely. When `contentLength` is `null`, the condition `contentLength && ...` is falsy and the guard is silently skipped. `response.text()` then reads the full body with **no enforced size limit**, defeating the `maxBytes` parameter.

**Fix** — enforce the limit during streaming:
```typescript
const chunks: Uint8Array[] = [];
let bytesRead = 0;
const reader = response.body?.getReader();
if (reader) {
  while (true) {
    const { done, value } = await reader.read();
    if (done) break;
    bytesRead += value.byteLength;
    if (bytesRead > maxBytes) {
      await reader.cancel();
      throw new Error(`File too large: exceeds ${maxBytes} bytes`);
    }
    chunks.push(value);
  }
}
const merged = new Uint8Array(bytesRead);
let offset = 0;
for (const chunk of chunks) { merged.set(chunk, offset); offset += chunk.length; }
return new TextDecoder().decode(merged);
```
The existing `Content-Length` header check can remain as an early-exit optimisation, but the stream-based check is required as the real guard.

shahariaazam (commented):
Thanks for the contribution. I found one issue that I just mentioned below. Would be happy to hear your thoughts on this.

src/lib/slack-client.ts
jiel: Good catch 👍 `Content-Length` is optional (chunked transfer, HTTP/2) so the guard was effectively a no-op in those cases.

Fixed: I kept the `Content-Length` check as an early-exit optimization and added the stream-based guard as the real enforcement, very close to your snippet. Only minor tweak is using `.byteLength` consistently in the merge loop (`.length` and `.byteLength` are identical on `Uint8Array`, but `.byteLength` is the canonical property for sized buffers).

shahariaazam (approved):

src/lib/slack-client.ts
shahariaazam: Thanks for the fix.

PR #36, opened 2026-04-11
fix: fetch all saved items and resolve thread replies

Sibirius:
## Summary
- Fixed `saved list` returning only 20 items instead of all saved items
- Fixed thread replies showing as "[message unavailable]" in saved items
- Added 10 unit tests for the saved item enrichment logic
- Updated `--state` help text to include `to_do` as a valid filter value

## Reason
The `saved list` command had a hardcoded default limit of 20 and used per-message `conversations.history` calls that can't resolve thread replies (only top-level messages). This meant users with more than 20 saved items only saw a subset, and any saved thread reply showed "[message unavailable]" even when the message was readable.

### Pagination
`saved.list` (browser auth) and `stars.list` (standard auth) both support cursor-based pagination. The command now pages through all results with a page size of 100, stopping early if `--limit` is specified. The `--limit` option no longer has a default value.

### Batch message resolution
The original code called `conversations.history` once per saved message to fetch its content. This was both slow (N+1 requests) and broken for thread replies, since `conversations.history` only returns top-level channel messages.

The fix replaces this with Slack's `messages.list` API, which accepts `{channel, timestamps[]}` groups and returns full message objects regardless of whether they're top-level messages or thread replies. Timestamps are grouped by channel and chunked into batches of 10 to avoid `too_many_channels` errors. Channel info is fetched in parallel with the message batches.

### Standard auth path
The `stars.list` API (standard auth) already includes message content inline, so no batch fetch is needed. This path is unchanged.

### Known limitations
- `messages.list` is an internal Slack API only available via browser auth. Standard auth continues to use `stars.list` which includes messages inline.

## QA
Tested manually against a workspace with 60+ saved items including top-level messages, thread replies, reminders (`to_do` state), and completed items. All items render correctly. Automated tests cover pagination, thread resolution, batching, limit truncation, both auth paths, and graceful fallbacks.

### DIFF
--- a/src/commands/saved.ts
+++ b/src/commands/saved.ts
@@ -11,8 +11,8 @@ export function createSavedCommand(): Command {
   saved
     .command('list')
     .description('List saved for later items')
-    .option('--limit <number>', 'Number of items to return', '20')
-    .option('--state <state>', 'Filter by state: saved or completed')
+    .option('--limit <number>', 'Maximum number of items to return')
+    .option('--state <state>', 'Filter by state: saved, to_do, or completed')
     .option('--workspace <id|name>', 'Workspace to use')
     .option('--json', 'Output in JSON format', false)
     .action(async (options) => {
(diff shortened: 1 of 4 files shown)


shahariaazam (approved):
Thanks for the fix.

PR #46, opened 2026-04-29
fix: preserve newlines in draft messages

### LINKED ISSUE #45: bug: draft messages lose newlines - parseMrkdwn creates multiple sections instead of preserving \n in text
## Bug Description
When creating a draft message with newlines via `slackcli messages draft`, the newlines are lost and the message appears as a single line in Slack's draft composer.

## Steps to Reproduce
```bash
slackcli messages draft \
  --recipient-id C1234567890 \
  --message $'Line 1\nLine 2\nLine 3'
```

## Expected Behavior
Draft should show:
```
Line 1
Line 2
Line 3
```

## Actual Behavior
Draft shows: `Line 1Line 2Line 3` (all on one line)

## Root Cause
In `src/lib/mrkdwn.ts`, `parseMrkdwn()` splits the input by `\n` and creates separate `rich_text_section` elements for each line:

```typescript
const lines = text.split('\n');
const sections: RichTextSection[] = lines.map(line => ({
  type: 'rich_text_section',
  elements: line.length > 0 ? parseInline(line) : [{ type: 'text', text: '\n' }],
}));
```

However, Slack's draft composer renders multiple `rich_text_section` elements inline, not as separate lines.

## Solution
Keep newlines embedded in a single `rich_text_section` with a single `text` element, rather than splitting into multiple sections.

Tested via curl: a draft with blocks `[{"type":"rich_text","elements":[{"type":"rich_text_section","elements":[{"type":"text","text":"Line1\nLine2\nLine3"}]}]}]` renders correctly with newlines.

## Comparison
- `slackcli messages send` works correctly (uses raw `text` param to `chat.postMessage`)
- `slackcli messages draft` fails (converts to blocks via `parseMrkdwn`)

## Environment
- slackcli version: 0.6.1
- OS: macOS
- Auth type: browser

heo001997:
## Summary
- Fixed `parseMrkdwn()` to preserve newlines embedded in text elements
- Slack's draft composer renders multiple `rich_text_section` elements inline, losing line breaks
- Changed to use a single section with `\n` characters in the text content
- Updated tests to match new behavior

## Reason
When creating a draft message with `slackcli messages draft --message $'Line1\nLine2'`, the newlines were lost and the message appeared as a single line. This was because `parseMrkdwn()` split text by `\n` into separate `rich_text_section` elements, but Slack's draft composer renders multiple sections inline without line breaks.

Fixes #45

### Root cause
The original implementation in `src/lib/mrkdwn.ts`:
```typescript
const lines = text.split('\n');
const sections = lines.map(line => ({
  type: 'rich_text_section',
  elements: parseInline(line),
}));
```

Slack's web composer renders each `rich_text_section` inline, not as separate lines.

### Solution
Keep newlines embedded in the text content within a single section:
```typescript
const elements = parseInline(text);
return [{
  type: 'rich_text',
  elements: [{
    type: 'rich_text_section',
    elements: elements,
  }],
}];
```

### Comparison
- `messages send` works correctly (uses raw `text` param to `chat.postMessage`)
- `messages draft` was broken (converted to blocks via `parseMrkdwn`)

Verified via direct curl to `drafts.create` that a single text element with embedded `\n` renders correctly in Slack's draft composer.

## QA
- [x] Draft messages now preserve newlines when viewed in Slack's composer
- [x] Inline formatting (*bold*, _italic_, `code`, ~strike~) still works
- [x] Newlines with mixed formatting preserved
- [x] `bun test src/lib/mrkdwn.test.ts` passes (16 tests)
- [x] Existing tests updated to match new behavior

### DIFF
--- a/src/lib/mrkdwn.test.ts
+++ b/src/lib/mrkdwn.test.ts
@@ -98,43 +98,42 @@ describe('parseMrkdwn', () => {
     }]);
   });
 
-  it('splits newlines into separate sections', () => {
+  it('preserves newlines in text elements', () => {
+    // Newlines must be embedded in text, not split into separate sections.
+    // Slack's draft composer renders multiple sections inline (no breaks),
+    // but correctly preserves \n within text elements.
     const result = parseMrkdwn('line one\nline two');
     expect(result).toEqual([{
       type: 'rich_text',
-      elements: [
-        {
-          type: 'rich_text_section',
-          elements: [{ type: 'text', text: 'line one' }],
-        },
-        {
-          type: 'rich_text_section',
-          elements: [{ type: 'text', text: 'line two' }],
-        },
-      ],
+      elements: [{
+        type: 'rich_text_section',
+        elements: [{ type: 'text', text: 'line one\nline two' }],
+      }],
     }]);
   });
 
-  it('handles formatting across multiple lines', () => {
-    const result = parseMrkdwn('*bold line*\n_italic line_');
+  it('preserves newlines with formatting', () => {
+    const result = parseMrkdwn('*bold*\nplain');
     expect(result).toEqual([{
       type: 'rich_text',
-      elements: [
-        {
-          type: 'rich_text_section',
-          elements: [{ type: 'text', text: 'bold line', style: { bold: true } }],
-        },
-        {
-          type: 'rich_text_section',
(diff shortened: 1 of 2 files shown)


shahariaazam (approved):

shahariaazam:
Thanks for the fix.

Issue #50, opened 2026-05-23
feat: add 'messages command' to execute slash commands and capture ephemeral replies (browser auth)


Eduard-Kuzhyr:
### Motivation

Slack slash commands like `/genie`, `/jira`, `/giphy`, and many internal team commands respond ephemerally — only the invoker sees the reply, and there is no public Web API to drive them. Today `slackcli` can send messages and add reactions but cannot execute a slash command or read what comes back, which blocks any AI agent or shell script that wants to interact with these flows.

Browser-auth (xoxc + xoxd) already gives us everything the Slack desktop client uses — the missing piece is wiring up the undocumented `chat.command` endpoint plus the MS WebSocket where the ephemeral reply arrives. With those in place we can offer a clean CLI verb that returns either rendered output or JSON.

### Proposal — CLI shape

```bash
slackcli messages command \
  --recipient-id <C... | U...> \
  --command /genie \
  [--text "args"] \
  [--timeout 15] \
  [--workspace <id|name>] \
  [--json]
```

- Requires browser auth. Standard `xoxb`/`xoxp` tokens get an early error pointing the user at `slackcli auth login-browser`.
- `--recipient-id` accepts either a channel ID or a user ID (DM is opened automatically — same convention as `messages send` / `messages draft`).
- Exit codes: `0` reply captured (or timed out but a synchronous body was returned), `1` execution / arg error, `2` no ephemeral reply within `--timeout`.
- `--json` emits `{ channel_id, command, text, timed_out, messages }` for scripting.
- Default human output renders the captured message(s) through the existing `formatMessage` pipeline so block-kit and legacy attachments show up in the terminal.

### Implementation notes

To keep the new command action small and testable, the work is split across a few focused modules:

- `src/lib/slack-client.ts` — `executeSlashCommand` (calls `chat.command` with `disp` + `client_token`) and `rtmConnect` (returns the `wss-primary.slack.com` URL plus the `Cookie`/`Origin`/`User-Agent` headers required for the WebSocket upgrade). The public `rtm.connect` endpoint returns a `LEGACY_BOT` URL that rejects browser sessions, hence the manual URL build.
- `src/lib/ephemeral-capture.ts` — pure helpers: `matchesEphemeral` (correlate frame to invocation via `is_ephemeral` / `subtype` / `client_msg_id` / self-targeted), `payloadToMessage`, `syncResponseToMessage`.
[...]

PR #51, opened 2026-05-23
feat: add 'messages command' for slash command capture

Eduard-Kuzhyr:
Adds a browser-auth subcommand that executes Slack slash commands via the chat.command endpoint and captures the ephemeral reply over the MS WebSocket. Splits the implementation across focused modules:

- block-renderer:        block-kit + legacy-attachment renderer extracted
                         from formatter so slash-command replies render
                         in the terminal
- ephemeral-capture:     pure helpers for matching frames and converting
                         payloads to SlackMessage
- slash-command-runner:  socket lifecycle orchestrator with single-cleanup
                         on every settle path (success, timeout, error,
                         invoke rejection)
- recipient:             resolveRecipientChannel helper used by send,
                         draft, and command (kills 3x duplication)

Also: validates --timeout input, parallelises captured-message user-info lookups via Promise.allSettled, and replaces the hardcoded UA version string with __APP_VERSION__.

Refs #50

### DIFF
--- a/src/commands/messages.ts
+++ b/src/commands/messages.ts
@@ -1,7 +1,10 @@
 import { Command } from 'commander';
 import ora from 'ora';
 import { getAuthenticatedClient } from '../lib/auth.ts';
-import { success, error } from '../lib/formatter.ts';
+import { success, error, formatMessage } from '../lib/formatter.ts';
+import { runSlashCommand } from '../lib/slash-command-runner.ts';
+import { resolveRecipientChannel } from '../lib/recipient.ts';
+import type { SlackUser } from '../types/index.ts';
 
 export function createMessagesCommand(): Command {
   const messages = new Command('messages')
@@ -22,13 +25,8 @@ export function createMessagesCommand(): Command {
       try {
         const client = await getAuthenticatedClient(options.workspace);
 
-        // Check if recipient is a user ID (starts with U) and needs DM opened
-        let channelId = options.recipientId;
-        if (options.recipientId.startsWith('U')) {
-          spinner.text = 'Opening direct message...';
-          const dmResponse = await client.openConversation(options.recipientId);
-          channelId = dmResponse.channel.id;
-        }
+        if (options.recipientId.startsWith('U')) spinner.text = 'Opening direct message...';
+        const channelId = await resolveRecipientChannel(client, options.recipientId);
 
         spinner.text = 'Sending message...';
         if (options.file) {
@@ -94,12 +92,8 @@ export function createMessagesCommand(): Command {
       try {
         const client = await getAuthenticatedClient(options.workspace);
 
-        let channelId = options.recipientId;
-        if (options.recipientId.startsWith('U')) {
-          spinner.text = 'Opening direct message...';
-          const dmResponse = await client.openConversation(options.recipientId);
-          channelId = dmResponse.channel.id;
-        }
+        if (options.recipientId.startsWith('U')) spinner.text = 'Opening direct message...';
+        const channelId = await resolveRecipientChannel(client, options.recipientId);
 
         spinner.text = 'Creating draft...';
         const response = await client.createDraft(channelId, options.message, {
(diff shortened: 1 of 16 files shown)


src/lib/slash-command-runner.ts
```diff
@@ -0,0 +1,103 @@
...
+        invoked = true;
+        try {
+          const syncResp = await opts.invokeCommand();
+          const synthetic = syncResponseToMessage(syncResp, opts.channelId);
+          if (synthetic) captured.push(synthetic);
+        } catch (e: any) {
+          settle(() => reject(e));
+          return;
+        }
+        timer = setTimeout(
```
shahariaazam: This timeout starts only after a `hello` frame. If the socket opens but Slack never sends `hello`, the promise never settles and the CLI hangs. Start a connection-level timeout immediately, or add a separate pre-hello timeout that calls `settle`.

src/lib/ephemeral-capture.ts
```diff
@@ -0,0 +1,63 @@
...
+  selfUserId?: string,
+): boolean {
+  if (!payload || payload.type !== 'message') return false;
+  if (payload.channel !== channelId) return false;
+
+  const isEphemeral = payload.is_ephemeral === true || payload.subtype === 'ephemeral';
+  const correlated = clientToken !== '' && payload.client_msg_id === clientToken;
+  const directlyTargeted = !!selfUserId && payload.user === selfUserId && payload.is_ephemeral === true;
+
+  return isEphemeral || correlated || directlyTargeted;
```
shahariaazam, line 34: This accepts any ephemeral message in the channel, so an unrelated ephemeral event can be captured as the command response. Require `client_msg_id === clientToken` when present, or add another invocation-specific correlation check before returning true.

shahariaazam (changes requested):
Thanks for the contribution. The structure and tests are solid, but two capture-path issues need fixing before this can merge.

src/lib/slash-command-runner.ts
Eduard-Kuzhyr: Fixed in 00fad2e.

Added an `armTimeout()` helper that arms a single timer at socket creation and re-arms it on the `hello` frame. Pre-hello waits are now bounded by `timeoutMs`; the post-invocation window stays exactly `timeoutMs`. Covered by a new `resolves with timedOut=true when hello never arrives` test in `slash-command-runner.test.ts`.

src/lib/ephemeral-capture.ts
Eduard-Kuzhyr: Fixed in 00fad2e.

Added a hard-reject in `matchesEphemeral` for frames whose `client_msg_id` belongs to another invocation (different non-empty value vs. our `clientToken`). Kept the heuristic fallback (`is_ephemeral` / `subtype: ephemeral` / self-target) because Slack does not stamp our `chat.command` `client_token` onto bot-emitted ephemerals — `client_token` is an idempotency key for our invocation, while `client_msg_id` is a Slack-side identifier on outbound user messages, so they live in different namespaces. An initial attempt to require strict equality always timed out against `/genie` in live testing; the conditional reject is the version that actually filters cross-invocation noise without dropping legitimate captures.

New tests: `rejects ephemeral whose client_msg_id belongs to another invocation` and `still rejects mismatched client_msg_id even with subtype=ephemeral`.

Eduard-Kuzhyr:
Follow-up on `f3050f7` — changed `runSlashCommand` semantics so `--timeout` is now a **collection window** instead of a first-match deadline.

**Why:** Slack's MS-socket protocol does not emit a terminator frame for slash commands. Commands like `/genie get <id>` send an immediate ack ephemeral and then the real payload several seconds later. The previous behaviour settled on the first match and dropped everything after it.

**Behaviour:**
- Default: collect every ephemeral for the full `--timeout` window, then return.
- New optional `--max-events <n>` flag: early-exit cap once N ephemerals have arrived.
- `timedOut` in JSON output now means "no ephemerals received" rather than "timer fired".

**Trade-off:** With `--max-events` unset, every successful call now blocks the full `--timeout`. Users control the wait by lowering `--timeout` or setting an explicit cap.

| Scenario | timeout | max-events | Result |
|---|---|---|---|
| `/genie` emits 3 ephemerals over 8s | 90 | unset | All 3 captured, returns at 90s. |
| Same | 90 | 3 | All 3 captured, returns immediately after the 3rd. |
| Command emits nothing | 15 | unset | `timedOut: true`, exit 2. |

Tests: 2 new (`without maxEvents collects every ephemeral until the timer fires`, `window stays open between matches: late frame still captured`), 2 existing pinned with `maxEvents: 1`, 1 assertion flipped to match the new `timedOut` semantic.

shahariaazam:
I need to test this PR as it's introducing a new feature. After testing I will get back soon @Eduard-Kuzhyr

Eduard-Kuzhyr:
Hey, @shahariaazam any updates on that?

shahariaazam (commented):
(superseded by newer review)

shahariaazam (changes requested):

Eduard-Kuzhyr:
@shahariaazam thanks for feedback, addressed mentioned issues.

src/lib/formatter.ts
```diff
@@ -103,11 +104,16 @@ export function formatMessage(
...
-  // Message text
-  const textLines = msg.text.split('\n');
-  textLines.forEach(line => {
-    output += `${indentStr}  ${line}\n`;
-  });
+  // Message text. Suppressed when blocks are present, since bots typically
+  // copy block content into `text` as a notification fallback — rendering
+  // both would duplicate the body.
+  const hasBlocks = !!(msg.blocks && msg.blocks.length > 0);
+  if (msg.text && !hasBlocks) {
```
shahariaazam, line 111: 🔴 **Blocker: text suppression drops message bodies**

This is my earlier suggestion implemented one step too broadly. It fires whenever `blocks.length > 0`, not when the blocks actually render anything.

**Why:** `block-renderer.ts` skips `image` and `actions`, and renders nothing for unknown types (`video`, `call`, `file`). `text` is already suppressed by then, so the body disappears. On `main` this path rendered `text` unconditionally, so this regresses `conversations read` (`conversations.ts:178` and `:257` reach `formatMessage` via `formatConversationHistory`).

Verified against the real function, ANSI stripped:

```
IMAGE-ONLY BLOCK, text: "Check out this chart!"
[11/14/2023, 22:13:20] @Alice
  ts: 1700000000.000100
                          <- body gone
```

Same for actions-only and `video` blocks. Any bot message carrying a chart, a button prompt or a newer block type loses its body. `formatter.test.ts:360` only covers a `section` block, which renders fine, so the suite misses this.

**Fix:** hoist the render, gate on its result.

```ts
const blockText = msg.blocks?.length ? renderBlocks(msg.blocks, indent + 2) : '';

// Suppress `text` only when blocks actually rendered a body; bots copy block
// content into `text` as a notification fallback.
if (msg.text && !blockText.trim()) {
  for (const line of msg.text.split('\n')) output += `${indentStr}  ${line}\n`;
}
```

Then reuse `blockText` at the existing render site instead of calling `renderBlocks` twice. Please add a test for an image-only block asserting the fallback text survives.

src/lib/block-renderer.ts
```diff
@@ -0,0 +1,130 @@
...
+      s += el.text || '';
+    } else if (el.type === 'link') {
+      s += el.text || el.url || '';
+    } else if (el.type === 'user') {
+      s += `<@${el.user_id}>`;
+    } else if (el.type === 'channel') {
+      s += `<#${el.channel_id}>`;
+    } else if (el.type === 'emoji') {
+      s += `:${el.name}:`;
+    }
```
shahariaazam, line 89: 🟠 **Major: rich_text drops `broadcast`, `usergroup` and `date` elements**

**Why:** `<!here> standup now` renders as ` standup now`, losing the @here entirely. This is independent of the blocker above: the block still returns non-empty text, so `text` stays suppressed and the mention is gone for good.

**Fix:** handle `broadcast` (`@${el.range}`), `usergroup` and `date`. Silently ignoring unrecognised element types means content vanishes with no signal.

src/commands/messages.ts
```diff
@@ -115,5 +109,107 @@ export function createMessagesCommand(): Command {
...
+
+        if (options.json) {
+          console.log(JSON.stringify({
+            channel_id: channelId,
+            command: options.command,
+            text: options.text || '',
+            timed_out: timedOut,
+            messages: captured,
+          }, null, 2));
+          return;
```
shahariaazam, line 189: 🟠 **Major: `--json` returns exit 0 on timeout**

**Why:** the non-JSON path exits 2 at line 194, this one returns 0. #50 specifies exit 2 for "no ephemeral reply within `--timeout`" with no `--json` carve-out. `--json` is the scripting path, so `slackcli messages command --json || handle_err` silently passes on a timeout.

**Fix:** apply the same `process.exit(2)` after emitting the JSON when `captured.length === 0`.

shahariaazam (changes requested):
## PR Review

**Verdict:** ❌ REQUEST CHANGES
**Scope check:** Matches #50's proposal (module split, CLI shape, tests). The `formatMessage` change reaches beyond it and regresses `conversations read`.

Earlier items verified fixed: pre-hello hang, socket `error` dropping captures, synthetic `ts` format. Type-check is clean and all 247 tests pass locally.

| # | Sev | Location | Issue | Action |
|---|-----|----------|-------|--------|
| 1 | 🔴 | `src/lib/formatter.ts:111` | Suppressing `text` when `blocks` exist drops the body whenever blocks render nothing (image, actions, unknown types) | Gate suppression on rendered output, not `blocks.length` |
| 2 | 🟠 | `src/lib/block-renderer.ts:79-89` | `flattenRichText` silently drops `broadcast`, `usergroup`, `date` | Handle those element types |
| 3 | 🟠 | `src/commands/messages.ts:189` | `--json` exits 0 on timeout, but #50 specifies exit 2 | Exit 2 after emitting JSON when nothing was captured |

Details inline.

**Housekeeping on my side, not blocking your work:**
- CI has not actually run on this head. `947380d8` and `a33e483d` are both `action_required`; the last green run was `f3050f70` (Jun 8), before the review-feedback commit and two `main` merges. I will approve the pending runs.
- #50 still has no `ready-for-pr` label. I will add it.
- The PR body says `Refs #50`, so merging will not auto-close the issue. Switch to `Closes #50` if you want that.

### To approve
Apply #1, #2 and #3 in this PR and I will re-review. The module split and the test suite are solid, and #1 is a narrow fix.

Eduard-Kuzhyr:
Closing this — no longer need it, won't have time to continue. Thanks for the work, feel free to reopen if useful elsewhere.

PR #76, opened 2026-07-15
fix: prevent --json output truncation at 64 KiB pipe buffer (#73)

### LINKED ISSUE #73: --json output truncated at ~64 KiB, producing invalid JSON (exit 0)
summary: Any command with `--json` whose response exceeds ~64 KiB has its stdout truncated mid-string, yielding unparseable JSON. The process still exits 0, so callers cannot distinguish truncation from success.

### Repro
Run any `--json` command that returns a large payload (e.g. a `search messages` page containing long messages):

```
slackcli search messages "<broad query>" --limit 20 --page 4 --sort timestamp --sort-dir asc --json | wc -c
# => 65600
```

The output is ~65600 bytes and is cut off inside a string, so parsing fails:

```
json.decoder.JSONDecodeError: Unterminated string starting at: line 1687 column 19 (char 65368)
```

Requesting the same item range with a smaller `--limit` (so each response stays under the cap) parses fine.

### Expected
Full, valid JSON regardless of payload size — or a non-zero exit code when the output is truncated, so scripts can detect it.

### Environment
- Binary: `slackcli-macos-arm64`
- Reproduced on **v0.6.1 and v0.7.0** (latest)
- The truncation boundary sits right around 64 KiB and slackcli ships as a Bun single-file executable, so this looks like stdout not being fully flushed/drained before process exit when writing to a non-TTY pipe, rather than per-command logic.

### Workaround
Paginate with a smaller `--limit` so each individual response stays under ~64 KiB.

shahariaazam:
Closes #73

## WHAT

Route every `--json` code path through a new `writeJson()` helper backed by `process.stdout.write`, instead of `console.log(JSON.stringify(...))`. 9 call sites across `search.ts`, `conversations.ts`, `canvas.ts` and `saved.ts`.

## WHY

Any `--json` command whose payload exceeds 64 KiB had its stdout truncated mid-string when piped, producing unparseable JSON, and still exited 0 so callers could not detect it.

Root cause is not the single-file executable, as the report hypothesised. `ora` reads `process.stdout.isTTY` at import time (via `cli-cursor` -> `restore-cursor@5.1.0`). Touching `process.stdout` materialises Bun's Node-compat `WriteStream`, which moves `console.log` off Bun's synchronous writer onto the async stream path. When stdout is a pipe, the first 64 KiB fills the OS pipe buffer and the remainder is queued, then the process exits before it drains. Every command starts an ora spinner, so every `--json` path was affected.

Evidence, 5.36 MB payload, real pinned tree (ora 8.2.0, restore-cursor 5.1.0):

| Variant | Bytes piped |
|---|---|
| `console.log` with ora | 65536 |
| `console.log` without ora | 5364912 |
| `process.stdout.isTTY` touched before the log | 65536 |
| `process.stdout.isTTY` touched after the log | 5364912 |
| `process.stdout.write()` with ora | 5364912 |
| `console.log` with ora, redirected to a file | 5364912 |

Pipe only. Neither `process.exit` nor `bun build --compile` is involved: plain `bun run` truncates identically, and the `--json` success paths never call `process.exit`.

## HOW

- `src/lib/formatter.ts`: add `writeJson(value)`, with a comment recording why `console.log` is unsafe here so the pattern does not regress.
- Replace all 9 `console.log(JSON.stringify(...))` sites with `writeJson(...)`.
- `src/lib/formatter.test.ts`: two tests. One pins the serialisation shape. The other is the regression test: it spawns a real child process with piped stdout, loads ora, and writes a >64 KiB payload. An in-process stub cannot catch this bug because it only manifests on a real pipe.

## Verification

- Type-check clean; 178 tests pass; `pre-commit run --all-files` fully green; build 92M, under the 150MB CI cap.
- Mutation-checked: reverting `writeJson` to `console.log` fails both new tests, so they genuinely pin the defect rather than passing vacuously.

## Deferred to #77

The same root cause affects stdout paths that are not `--json`, all tracked in #77:

- `canvas read --raw` (`canvas.ts:144`) writes up to 10 MB of HTML via `console.log`. Machine-consumable, so this is the more urgent half.
- Human-readable output: `formatSearchMessages` (`search.ts:71`) and `formatConversationHistory` (`conversations.ts:178,257`) truncate identically when piped to `less`.

Kept out of this PR because #73 is specifically about `--json`, and CLAUDE.md asks changes to stay focused. #77 carries the rest with a `writeOut()` sibling helper.

### DIFF
--- a/src/commands/canvas.ts
+++ b/src/commands/canvas.ts
@@ -1,7 +1,7 @@
 import { Command } from 'commander';
 import ora from 'ora';
 import { getAuthenticatedClient } from '../lib/auth.ts';
-import { error, formatCanvasList, formatCanvasContent } from '../lib/formatter.ts';
+import { error, formatCanvasList, formatCanvasContent, writeJson } from '../lib/formatter.ts';
 import { canvasHtmlToMarkdown, isAuthPage } from '../lib/canvas-parser.ts';
 import type { SlackCanvas, SlackUser } from '../types/index.ts';
 
@@ -48,7 +48,7 @@ export function createCanvasCommand(): Command {
         spinner.succeed(`Found ${files.length} canvases`);
 
         if (options.json) {
-          console.log(JSON.stringify({
+          writeJson({
             canvas_count: files.length,
             canvases: files.map(f => ({
               id: f.id,
@@ -60,7 +60,7 @@ export function createCanvasCommand(): Command {
               size: f.size,
               permalink: f.permalink,
             })),
-          }, null, 2));
+          });
           return;
         }
 
@@ -201,7 +201,7 @@ export function createCanvasCommand(): Command {
         }
 
         if (options.json) {
-          console.log(JSON.stringify({
+          writeJson({
             id: file.id,
             title: file.title || file.name,
             created: file.created,
@@ -211,7 +211,7 @@ export function createCanvasCommand(): Command {
             size: file.size,
             permalink: file.permalink,
             markdown,
-          }, null, 2));
+          });
           return;
         }
 
--- a/src/commands/conversations.ts
+++ b/src/commands/conversations.ts
@@ -2,7 +2,7 @@ import chalk from 'chalk';
 import { Command } from 'commander';
 import ora from 'ora';
 import { getAuthenticatedClient } from '../lib/auth.ts';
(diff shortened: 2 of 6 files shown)


shahariaazam:
Deferred to #77: `canvas read --raw` (`canvas.ts:144`) and the human-readable output paths share #73's root cause but fall outside its `--json` wording. Tracked there rather than widening this PR.

PR #80, opened 2026-07-19
:bug: fix: stop hardcoding stale version in local bun builds

### LINKED ISSUE #79: 🐛 Local bun builds show stale version / false update warning
Local builds hardcode `__APP_VERSION__` to `0.2.2` in package.json scripts, so running from current source still reports an old version and nags about updates (latest is much newer).

Running via `bun` from source has the same problem; version isn't taken from package.json.

a-tokyo:
Fixes #79

Local bun builds were baking in `0.2.2`, so even current source looked outdated and showed an update warning.

Now the version comes from package.json, and we skip the update nag when running under bun from source.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,16 +1,16 @@
 {
   "name": "slackcli",
-  "version": "0.2.0",
+  "version": "0.7.0",
   "description": "A fast, developer-friendly CLI tool for interacting with Slack workspaces",
   "main": "src/index.ts",
   "type": "module",
   "scripts": {
     "dev": "bun run src/index.ts",
-    "build": "bun build --compile --minify --sourcemap --define __APP_VERSION__='\"0.2.2\"' src/index.ts --outfile dist/slackcli",
+    "build": "bun run scripts/build.ts",
     "build:all": "bun run build:linux && bun run build:macos && bun run build:windows",
-    "build:linux": "bun build --compile --minify --target=bun-linux-x64 --define __APP_VERSION__='\"0.2.2\"' src/index.ts --outfile dist/slackcli-linux",
-    "build:macos": "bun build --compile --minify --target=bun-darwin-x64 --define __APP_VERSION__='\"0.2.2\"' src/index.ts --outfile dist/slackcli-macos",
-    "build:windows": "bun build --compile --minify --target=bun-windows-x64 --define __APP_VERSION__='\"0.2.2\"' src/index.ts --outfile dist/slackcli-windows.exe",
+    "build:linux": "bun run scripts/build.ts --target=bun-linux-x64 --outfile=dist/slackcli-linux",
+    "build:macos": "bun run scripts/build.ts --target=bun-darwin-x64 --outfile=dist/slackcli-macos",
+    "build:windows": "bun run scripts/build.ts --target=bun-windows-x64 --outfile=dist/slackcli-windows.exe",
     "test": "bun test",
     "type-check": "bunx tsc --noEmit"
   },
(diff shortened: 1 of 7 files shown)


src/lib/updater.ts
```diff
@@ -212,6 +212,11 @@ export function getUpdateCommand(): string {
 // Show a one-line update notification after the command finishes (via beforeExit),
 // and refresh the cache in the background if it is stale.
 export function notifyIfUpdateAvailable(): void {
+  // Local `bun run` / source checkout — not a release binary; skip self-update nags.
+  if (isRunningUnderBun()) {
```
vibexp-agent, line 221: 🔴 **This guard also needs to be in `performUpdate()`.** Before this PR, source runs had `CURRENT_VERSION = 'dev'`, so `isNewerVersion()` always returned `false` and self-update was inert. Now `CURRENT_VERSION` is a real semver — once the next release ships, running `slackcli update` under `bun run` will download the binary and `rename()` it over `process.execPath`, which under Bun is the **Bun interpreter itself** (`~/.bun/bin/bun`), destroying the user's Bun install.

Add the same `isRunningUnderBun()` check at the top of `performUpdate()` and bail with a "running from source — use `git pull`" message.

package.json
```diff
@@ -1,16 +1,16 @@
 {
   "name": "slackcli",
-  "version": "0.2.0",
+  "version": "0.7.0",
```
vibexp-agent, line 3: 🟡 This will silently drift again: `release.yml` derives the version from the git tag and never syncs/verifies `package.json`, so on the next release this is stale once more. Consider a release-time check that the tag matches `package.json` version (fine as a follow-up issue).

scripts/build.ts
```diff
@@ -0,0 +1,31 @@
...
+const extraArgs = process.argv.slice(2).filter((a) => !a.startsWith('--outfile='));
+const cwd = join(import.meta.dir, '..');
+
+const result = Bun.spawnSync(
+  [
+    'bun',
+    'build',
+    '--compile',
+    '--minify',
+    '--sourcemap',
```
vibexp-agent: 🔵 Nit: `--sourcemap` is now applied to `build:linux/macos/windows` too, which the old commands didn't have (release CI is unaffected since it invokes `bun build` directly). Intentional?

vibexp-gh (changes requested):

src/lib/updater.ts
a-tokyo: Added the same `isRunningUnderBun()` bail at the top of `performUpdate()`, so a source run points you at `git pull` instead of renaming over the Bun binary.

scripts/build.ts
a-tokyo: only the local `build` gets `--sourcemap`; the cross-platform targets stay without it

package.json
a-tokyo: Agreed, it'll drift again since releases bake the tag rather than package.json. Opened #81 to add a release-time tag↔package.json check as a follow-up so this PR stays focused.

a-tokyo:
Addressed the review notes; Bun guard on `performUpdate()`, sourcemap restored for local builds only, and opened #81 for the package.json drift follow-up.

vibexp-gh (approved):
Thanks @a-tokyo for the clean fix and the quick turnaround on the review feedback — the `performUpdate()` guard, the sourcemap scoping, and opening #81 for the version-drift follow-up. CI is green, tests and type-check pass locally. 🚀

PR #91, opened 2026-08-01
:sparkles: feat(auth): add login-auto for automated browser token capture

### LINKED ISSUE #90: ✨ feat: auth login-auto — sign in through a browser instead of copying tokens by hand
## What

A new command, `slackcli auth login-auto`. It opens a browser, waits for you to sign into Slack, and captures the `xoxd`/`xoxc` pair itself. One sign-in covers every workspace you're signed into.

Existing commands stay as they are.

## Why

Browser tokens are the "no Slack app needed" path, and right now they're the fiddliest part of the CLI. You open DevTools, hunt for a Slack API request, copy one token out of the Cookie header and another out of the request payload. `auth parse-curl` takes some of the sting out, but you still have to go into DevTools, and you do it again for every workspace.

It's also easy to get wrong. The `xoxd` value is percent-encoded. `xoxc` turns up in three different body encodings depending on the endpoint. Pick the wrong request and you get a token that authenticates nothing. `curl-parser.ts` is as big as it is for exactly these reasons.

Reading the tokens straight out of a signed-in browser removes that whole class of mistake, and picks up multi-workspace users in one pass.

## How (a suggestion)

Drive an already-installed Chrome, Edge, Chromium or Brave over the DevTools Protocol:

- `Network.requestWillBeSent` to read `xoxc` out of Slack API request bodies. Handles multipart, JSON and urlencoded, since all three show up in normal traffic.
- `Storage.getCookies` for the `d` cookie. This is the reason browser automation is needed at all, the cookie is HttpOnly so page JavaScript can't touch it.
- `localConfig_v2` in localStorage to pick up the remaining workspaces. Best effort, so a shape change costs you the extra workspaces rather than the run.

Verification and storage go through the existing `authenticateBrowser()`, so a workspace added this way is indistinguishable from one added by hand.

**No new dependency.** Playwright is the obvious tool, but it doesn't fit how slackcli ships. Bundling it into the compiled binary blows the 150MB CI limit, and an external Playwright can't resolve from a downloaded binary at all, so the feature would quietly only work from source. CDP over Bun's built-in WebSocket has neither problem and leaves the binary size unchanged.

**A dedicated browser profile is required, not just tidier.** Since Chrome 136 the browser refuses `--remote-debugging-port` when pointed at the default user-data directory. The cost is one sign-in inside the slackcli profile; it persists, so later runs need no interaction.

Happy to open a PR if this gets triaged and picks up `ready-for-pr`.

a-tokyo:
## Summary

Adds `slackcli auth login-auto`. It opens a browser, you sign into Slack as normal, and the CLI reads the `xoxd`/`xoxc` pair itself. No DevTools, no copy-paste, and one sign-in enrols every workspace you're signed into.

```bash
slackcli auth login-auto
```

Closes #90

## Changes

- `src/lib/cdp-client.ts` — a small Chrome DevTools Protocol client over Bun's built-in WebSocket. Correlates replies by id, dispatches events, times commands out, and rejects anything in flight if the socket closes so a browser you quit halfway through reports instead of hanging.
- `src/lib/browser-launcher.ts` — finds Chrome, Edge, Chromium or Brave (with a `SLACKCLI_BROWSER` override) and launches it against a dedicated profile with debugging on an OS-assigned port.
- `src/lib/browser-auth.ts` — the capture logic, plus parsers for the three request encodings Slack uses, the percent-encoded `d` cookie, and `localConfig_v2`.
- `authenticateAuto()` in `src/lib/auth.ts` — routes verification and storage through the existing `authenticateBrowser()`, so a workspace added this way is identical to one added by hand. If one workspace fails it's reported and the rest still save.

## No new dependency

Playwright would be the obvious choice and it doesn't fit. slackcli ships a `bun build --compile` binary: bundling Playwright exceeds the 150MB CI limit, and an external one can't resolve from a downloaded binary, so the command would work from source and quietly break for everyone else. CDP over Bun's WebSocket avoids both. Binary measured before and after: 64,304,738 → 64,337,762 bytes, about 32KB.

## Browser profile

The browser runs against `~/.config/slackcli/browser-profile`, never your own. Since Chrome 136 the browser refuses `--remote-debugging-port` when pointed at the default user-data directory, so this isn't a preference. You sign in once inside that profile; it persists, and later runs need no interaction.

`auth logout` deletes the profile too, since while it exists `login-auto` can mint fresh tokens without prompting. `--keep-browser-session` opts out. Deletion only ever touches a directory slackcli created, checked by a marker file and `lstat` so a symlink or someone else's directory can't be removed by mistake.

## Security

- No token value is printed on any `login-auto` path.
- Workspace URLs are gated to `https://` on a `slack.com` host before a session cookie is sent anywhere. The URL comes out of the page's own localStorage, so it can't be trusted on its own.
- `--workspace-url` is validated at the command layer and again before exec, so a value starting with `-` can't become a browser switch.
- Profile directory is `0700`; `workspaces.json` stays `0600`.
- While the browser is open it has a loopback DevTools port that any local process could connect to. The browser is closed as soon as capture finishes, and on Ctrl-C, SIGTERM or SIGHUP. Documented in the README.

## Tests

104 new tests, 290 total. Covers all three token encodings, look-alike cookie domains, hostile URLs planted in localStorage, argument injection, Windows browser resolution, and the profile-deletion guards.

Verified against a real workspace from the compiled binary: fresh tokens captured, 98 conversations fetched, re-login 5s headless with no interaction, no browser processes left behind, no token in any output.

## Not covered

- Multi-workspace enrolment is implemented and unit-tested but I only had one workspace to try it against.
- Windows and Linux browser resolution is unit-tested with injected platforms, not run on those machines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,5 +54,3 @@ secrets/
 credentials/
 *.secret
 
-# Tribunal verification scratch (not a deliverable)
-.tribunal/
--- a/.gitignore
+++ b/.gitignore
@@ -53,4 +53,3 @@ docs/internal/
 secrets/
 credentials/
 *.secret
-
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,6 @@ docs/internal/
 secrets/
 credentials/
 *.secret
+
+# Tribunal verification scratch (not a deliverable)
+.tribunal/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Drives an already-installed Chrome/Edge/Chromium/Brave over the Chrome DevTools Protocol; **no new runtime dependency** and no change to binary size
   - Uses a dedicated browser profile (required: Chrome 136+ refuses remote debugging against the default profile), so only the first run needs interaction
   - `--workspace-url`, `--timeout`, `--headless`; `SLACKCLI_BROWSER` and `SLACKCLI_BROWSER_PROFILE` env overrides
-  - Typed failure reasons (browser not found, launch timeout, capture timeout, missing cookie) each with actionable guidance
+  - Typed failure reasons (browser not found, launch timeout, capture timeout, browser closed, missing cookie) each with actionable guidance
+  - Workspace URLs are gated to `https://` on a `slack.com` host before any session cookie is sent to them
+
+### Changed
+- `auth logout` now also deletes the `login-auto` browser profile, which is a credential store in its own right — while it exists, `login-auto` re-mints working tokens without prompting. Use `--keep-browser-session` to retain the old behaviour.
 
 ## [0.2.0] - 2026-01-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Typed failure reasons (browser not found, launch timeout, capture timeout, browser closed, missing cookie) each with actionable guidance
   - Workspace URLs are gated to `https://` on a `slack.com` host before any session cookie is sent to them
 
+### Fixed
+- `login-auto` now asks Chrome to close itself (CDP `Browser.close`) before falling back to signals, then sweeps any helper the browser leaves behind — previously each run stranded ~5 Chrome processes. Order matters: sweeping *instead of* a clean shutdown leaves the profile unopenable.
+
 ### Changed
 - `auth logout` now also deletes the `login-auto` browser profile, which is a credential store in its own right — while it exists, `login-auto` re-mints working tokens without prompting. Use `--keep-browser-session` to retain the old behaviour.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- **Automatic browser login** (`auth login-auto`): sign into Slack in a browser and let SlackCLI capture the `xoxd`/`xoxc` tokens — no DevTools, no copy-paste
+  - Enrols every workspace the user is signed into from a single sign-in
+  - Drives an already-installed Chrome/Edge/Chromium/Brave over the Chrome DevTools Protocol; **no new runtime dependency** and no change to binary size
+  - Uses a dedicated browser profile (required: Chrome 136+ refuses remote debugging against the default profile), so only the first run needs interaction
+  - `--workspace-url`, `--timeout`, `--headless`; `SLACKCLI_BROWSER` and `SLACKCLI_BROWSER_PROFILE` env overrides
+  - Typed failure reasons (browser not found, launch timeout, capture timeout, missing cookie) each with actionable guidance
+
 ## [0.2.0] - 2026-01-30
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ A fast, developer-friendly command-line interface tool for interacting with Slac
 ## Features
 
 - 🔐 **Dual Authentication Support**: Standard Slack tokens (xoxb/xoxp) or browser tokens (xoxd/xoxc)
+- 🪄 **Automatic Browser Login**: `auth login-auto` — sign into Slack in a browser and the tokens are captured for you, no DevTools, no copy-paste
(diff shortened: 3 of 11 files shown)


vibexp-gh (dismissed):
Reviewed against the head commit. Verified locally: `tsc --noEmit` clean, `bun test` 290/290 pass. CI green (type-check → build → 150MB binary check).

**Approving.** No blocking issues found.

### What I checked

- **Auth path is not duplicated.** `authenticateAuto()` routes verification and persistence through the existing `authenticateBrowser()`, so `testAuth()` remains the single place a token is judged valid and a workspace enrolled this way is byte-identical to one from `login-browser` / `parse-curl`. This was the main thing I wanted to confirm and it holds.
- **URL gating.** `isSlackWorkspaceUrl()` anchors on a dot boundary and rejects `app.slack.com`; the check is applied again in `authenticateAuto()` immediately before the session cookie travels. Correctly treated as load-bearing rather than validation hygiene.
- **Profile deletion.** The sentinel + `lstat` ownership guard (regular file, not directory; no symlink follow) means `SLACKCLI_BROWSER_PROFILE` pointed at a real browser profile can't be recursively deleted by `logout`, and `claimProfileDir()` refuses to stamp a directory it didn't create. That's the right shape for a recursive delete driven by user input.
- **Argument injection.** `--workspace-url` validated at the command layer and again in `isSafeStartUrl()` before exec. Leading-`-` rejection is the correct last gate for positional argv.
- **Shutdown.** `Browser.close` → SIGTERM → SIGKILL → profile-matched sweep, with the ordering rationale documented. Registering the reaper immediately after `spawn` rather than after CDP attach is the right call — that window is exactly when a user interrupts.

### Non-blocking notes for a follow-up

1. `sweepProfileHelpers()` interpolates `profileDir` into a `pkill -f` pattern, which pkill reads as an ERE. Regex metacharacters in a custom `SLACKCLI_BROWSER_PROFILE` would widen the match beyond this browser's helpers. Mitigated in practice because `claimProfileDir()` has already established the path is ours, and it's self-inflicted via an env var the user sets — but escaping the pattern would close it.
2. `openBrowserSession()` declares an `onProgress` option it never calls. Dead parameter; either wire it to the launch phase or drop it.
3. `--timeout` uses `parseInt`, so `300abc` silently becomes `300`. Cosmetic.
4. `--disable-blink-features=AutomationControlled` is a deliberate automation-detection choice. Consistent with what this CLI already is, but flagging it so it's an explicit decision rather than an inherited one.

### One thing to watch

`localConfig_v2` is undocumented Slack client state and the Chrome launch-flag surface moves. The v2→v3 profile-format bump inside this PR is a preview of that maintenance cost. The degradation path is right (a localStorage shape change costs the extra workspaces, not the run), but expect this file to need occasional attention that the test suite can't warn you about.

Thanks for the thorough write-up on why not Playwright, and for the measured before/after binary size — that made the dependency argument easy to check.

vibexp-agent:
@a-tokyo — this is approved and green, but the merge is blocked by one last thing that isn't about the code.

`main` has a **require signed commits** rule, and 7 of the 8 commits here are unsigned (only `951083b` is verified). No maintainer can bypass it — the ruleset has no bypass actors — so the commits themselves need signatures.

If you already have signing configured:

```bash
git fetch origin
git rebase --exec 'git commit --amend --no-edit -S' origin/main
git push --force-with-lease
```

If you don't, SSH signing is the quickest route (uses a key you likely already have on your GitHub account):

```bash
git config --global gpg.format ssh
git config --global user.signingkey ~/.ssh/id_ed25519.pub
git config --global commit.gpgsign true
```

Then run the rebase above. Make sure that key is added as a **signing key** in GitHub → Settings → SSH and GPG keys (an authentication key is a separate entry, even for the same key).

Once the commits show as *Verified*, I'll squash-merge it — nothing else is outstanding. Sorry for the extra hoop, and thanks for the unusually thorough write-up on this one.

a-tokyo:
@vibexp-gh - Done

a-tokyo:
Followups done.

vibexp-gh (approved):
Re-reviewed at `cc8a2b5`. All 11 commits now verified, so the ruleset is satisfied.

The force-push carried more than signatures — I diffed the new head against the commit I approved previously (`c0db013`) to be sure. Three of the four non-blocking notes were addressed:

- `escapeEre()` now escapes the profile path before it reaches `pkill -f`. Character class covers `\` itself, so the escaping is not self-defeating.
- `openBrowserSession()` no longer declares the `onProgress` option it never used.
- `--timeout` uses `Number()` + `Number.isInteger()`, so `300abc` is now rejected instead of silently becoming `300`.

The fourth (`--disable-blink-features=AutomationControlled`) was a decision to make consciously rather than a defect, and it stands.

Verified locally on this head: `tsc --noEmit` clean, `bun test` 292/292 pass (up from 290 — the two new ones cover the ERE escaping). CI green.

Approving again. Merging now.

PR #93, opened 2026-08-04
Add nix flake for nix packaging and development

Flexksx:


vibexp-agent (commented):
Thanks for this, and sorry it sat unreviewed for so long. That was our fault, not yours: our automated reviewer only started running on 20 August, a day after this PR had already been marked stale, and it skips anything carrying that label. So it never saw this. We have fixed the gap; this review is by hand.

The packaging itself reads well. `nix/package.nix` deriving `pname` and `version` from `package.json` rather than restating them is the right call, the fileset is tight, and generating `nix/bun.nix` from `bun.lock` with a CI check that regenerates and diffs it is a better answer than most flakes ship with. I checked that `bun run build` with no arguments writes `dist/slackcli`, which is what `installPhase` installs, and that `scripts/build.ts` takes `__APP_VERSION__` from `package.json` — so the derivation's version and `slackcli --version` agree.

Two things worth looking at.

**1. Intel macOS is dropped.** `systems.url = "github:nix-systems/triplet"` gives `x86_64-linux`, `aarch64-linux` and `aarch64-darwin` — which your `nix/README.md` states accurately. But `release.yml` publishes a `bun-darwin-x64` binary, so Intel Mac is currently a supported platform for every other install method and not for nix. If that is deliberate it is worth a line in `nix/README.md` saying so; if not, `nix-systems/default` covers `x86_64-darwin` too.

**2. Regenerating `nix/bun.nix` needs nix.** Any change to `bun.lock` turns the Nix check red, and the only documented fix is `nix develop -c bun2nix -o nix/bun.nix`, which a contributor without nix installed cannot run. That is a reasonable trade, but it does mean a dependency bump can block on someone having a nix toolchain. Worth a sentence in `CONTRIBUTING.md` pointing at `nix/README.md`, so whoever hits it knows what happened.

Neither is a correctness problem and I found nothing that breaks.

Two notes on process rather than code, so you know where this stands: no workflow run has executed on this PR yet — fork PRs need a maintainer to approve them — so nothing here has been verified by CI, including the new `nix.yml` itself. And merging it is @shahariaazam's call, not mine; I only review.

Flexksx:
Thanks for the thorough review and the kind note about the delay.

Both points addressed in [0ed21c0](https://github.com/Flexksx/slackcli/commit/0ed21c0):

1. **Intel macOS**: switched `nix-systems/triplet` to `nix-systems/default`, which adds `x86_64-darwin`. Updated `nix/README.md` to list all four systems.

2. **bun.nix regeneration**: added a "Nix packaging" section to `CONTRIBUTING.md` explaining the regeneration step and its Nix requirement, with a pointer to `nix/README.md`.

Investigated alternatives to the checked-in `bun.nix` (fixed-output derivation via nixpkgs `bun.fetchDeps`, npm-based `bun2nix`). The FOD approach is still in draft in nixpkgs with reproducibility issues, and keeping `bun2nix` in the Nix dev shell is the cleaner fit for this project. The doc note is the pragmatic answer for now.

Ready for CI approval whenever a maintainer can trigger it.

nix/README.md
```diff
@@ -0,0 +1,92 @@
...
+## Run it without installing
+
+```bash
+nix run github:shaharia-lab/slackcli -- --help
+```
+
+Pin to a release or a commit by appending a git ref:
+
+```bash
+nix run github:shaharia-lab/slackcli/v0.7.1 -- --version
```
vibexp-agent, line 23: This example fails today. `v0.7.1` is a real tag but it predates the flake, and so does every other tag up to `v0.9.1`, so `nix run github:shaharia-lab/slackcli/v0.7.1` errors with `flake.nix does not exist`. Worth using a placeholder like `github:shaharia-lab/slackcli/<tag>` until a release actually ships the flake.

vibexp-agent (changes requested):
Both points from the last round check out: `nix-systems/default` at the pinned rev gives exactly the four systems `nix/README.md` now lists, and the CONTRIBUTING note covers the regeneration step. One small thing left, inline.

vibexp-agent:
Hi @Flexksx, thanks for the time you put into this.

Stepping back to where this actually stands.

This repository requires every pull request to be linked to an open GitHub issue that has been triaged and carries the `ready-for-pr` label ([CONTRIBUTING.md](https://github.com/shaharia-lab/slackcli/blob/main/CONTRIBUTING.md)). That rule exists precisely for a change like this one. There is no linked issue here and the PR description is empty, so neither the community nor I have anything to evaluate the change against, and there is no record of the discussion that should have happened before the code was written.

The most useful thing you can do is open a separate issue covering:

- **WHAT** the change is: a flake exposing the CLI as a package and a dev shell, with CI to keep the generated dependency set honest.
- **WHY** it is needed: who is asking for it, what installing slackcli on NixOS looks like today without it, and why the existing install paths (release binaries, Homebrew, from source) are not enough.
- **HOW** (optional, and mostly already answered by this PR).

We can then discuss it in the open and decide whether it is worth marking `ready-for-pr`.

Being straight with you about the current view: we do not see an immediate need for nix packaging in the core project. It adds a flake, a lockfile, a generated `bun.nix` and a CI job that all have to be kept working on every dependency change, and so far nobody has asked for it. So I am labelling this `wontfix` for now. That is about need and maintenance cost, not about the quality of the code. If an issue is opened and gets triaged as ready, we will revisit this PR.

Thanks again, and sorry the first response you got was a review rather than this conversation.

PR #97, opened 2026-08-06
fix(auth): parse cURL commands that pass the URL via --url flag

### LINKED ISSUE #94: auth parse-curl fails on Chrome 151+ "Copy as cURL" output (URL passed via --url flag)
## What

`slackcli auth parse-curl` (including `auth parse-curl --from-clipboard`) fails with:

```
❌ Error: Failed to parse cURL command
   Could not find Slack workspace URL in cURL command
```

…when the pasted cURL was produced by **Chrome 151+ DevTools → "Copy as cURL"**, because Chrome now prefixes the request URL with the `--url` flag:

```
curl --url 'https://<workspace>.slack.com/api/...' \
  -H '...' \
  -b '...; d=xoxd-...' \
  --data-raw $'------WebKitFormBoundary...name="token"...xoxc-...'
```

instead of the older **positional** form:

```
curl 'https://<workspace>.slack.com/api/...' \
  ...
```

## Why

Chrome 151 changed "Copy as cURL" to *"prefix target URLs with `--url` to prevent command-line flag misinterpretation for dash-prefixed URLs"* — see the official release notes: https://developer.chrome.com/blog/new-in-devtools-151

The parser's workspace-URL regex in `src/lib/curl-parser.ts` is anchored to the URL appearing **immediately after** `curl` (`/curl\s+'?(https?:\/\/([\w.-]+)\.slack\.com...)/`). With the new format, the `--url ` token sits between `curl` and the URL, so the match fails and parsing aborts before it ever looks at the tokens.

Notably, everything else in the new format already parses correctly — the `xoxc` token in the multipart `--data-raw` (`name="token"`) and the `xoxd` cookie in `-b` are both extracted fine once the URL is found. **The `--url` prefix is the sole blocker.** (Verified: manually removing `--url ` so the URL is positional makes the exact same command parse successfully.)

This breaks the primary documented onboarding path (Copy as cURL → `auth parse-curl`) for anyone on current Chrome.

## Steps to reproduce

1. Chrome 151+ → open a Slack tab → DevTools → Network.
2. Right-click any `slack.com/api/...` request → Copy → **Copy as cURL**.
3. Run `slackcli auth parse-curl --from-clipboard`.
4. Observe `Could not find Slack workspace URL in cURL command`.

## How (suggested fix)

Allow the workspace-URL regex to accept the URL when it follows the `--url` flag as well as positionally, e.g.:

```ts
const urlMatch = curlInput.match(
  /(?:curl\s+|--url(?:\s+|=))['"]?(https?:\/\/([\w.-]+)\.slack\.com[^'"\s]*)/
);
```

Plus a `curl-parser.test.ts` case for the Chrome 151 `--url` format. Small, focused change confined to `src/lib/curl-parser.ts` and its test.

## Environment

- slackcli 0.7.1 (Homebrew, macOS arm64)
- Google Chrome 151

vibexp-gh:
Fixes #94

## What

`auth parse-curl` failed with `Could not find Slack workspace URL in cURL command` for cURL copied from Chrome 151+ DevTools, which now prefixes the target URL with the `--url` flag instead of passing it positionally.

The workspace-URL regex was anchored to the URL appearing immediately after `curl`. It now accepts the URL either positionally or after `--url` (with a space or `=` separator). Everything else in the new format already parsed correctly, so this is the only change needed.

## Tests

Added `curl-parser.test.ts` cases for the Chrome 151 `--url 'https://…'` format and the `--url="https://…"` variant (enterprise domain), and included both in the real-world formats loop.

`bun test` 188 pass / 0 fail, `bun run type-check` clean.

### DIFF
--- a/src/lib/curl-parser.test.ts
+++ b/src/lib/curl-parser.test.ts
@@ -47,7 +47,8 @@ const CURL_CHROME_151_URL_FLAG = `curl --url 'https://chrometeam.slack.com/api/c
   -H 'origin: https://app.slack.com' \\
   --data-raw $'------WebKitFormBoundary\\r\\nContent-Disposition: form-data; name="token"\\r\\n\\r\\nxoxc-chrome151-test\\r\\n------WebKitFormBoundary--\\r\\n'`;
 
-// --url with an = separator, and an enterprise domain
+// curl itself rejects --url=value, but hand-edited commands use it — accept it defensively.
+// Enterprise domain, double-quoted URL.
 const CURL_URL_FLAG_EQUALS = `curl --url="https://chromeorg.enterprise.slack.com/api/users.list" \\
   -b 'd=xoxd-url-equals-token' \\
   --data-raw $'------Boundary\\r\\nContent-Disposition: form-data; name="token"\\r\\n\\r\\nxoxc-url-equals-test\\r\\n------Boundary--\\r\\n'`;
@@ -83,10 +84,16 @@ describe('parseCurlCommand', () => {
       expect(result.workspaceUrl).toBe('https://chrometeam.slack.com');
     });
 
-    it('should extract workspace from --url=value format', () => {
+    it('should extract workspace from the --url=value spelling', () => {
       const result = parseCurlCommand(CURL_URL_FLAG_EQUALS);
       expect(result.workspaceName).toBe('chromeorg');
       expect(result.workspaceUrl).toBe('https://chromeorg.enterprise.slack.com');
+    });
+
+    it('should prefer the --url target over a Slack URL in an earlier header', () => {
+      const curlWithDecoyHeader = `curl -H 'referer: https://decoyteam.slack.com/' --url 'https://realteam.slack.com/api/test' -b 'd=xoxd-decoy-test' --data-raw '{"token":"xoxc-decoy-test"}'`;
+      const result = parseCurlCommand(curlWithDecoyHeader);
+      expect(result.workspaceName).toBe('realteam');
     });
 
     it('should throw CurlParseError for missing workspace URL', () => {
--- a/src/lib/curl-parser.test.ts
+++ b/src/lib/curl-parser.test.ts
@@ -40,6 +40,18 @@ const CURL_JSON_BODY = `curl 'https://jsonworkspace.slack.com/api/some.method' \
   -b 'd=xoxd-json-test-token' \\
   --data-raw '{"token":"xoxc-json-test-111-222","as_admin":false}'`;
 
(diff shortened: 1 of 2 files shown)


shahariaazam (approved):
## PR Overview

- **Title:** fix(auth): parse cURL commands that pass the URL via `--url` flag
- **Scope:** 2 files changed, +38/-1
- **Category:** Bug fix
- **Risk Level:** Low
- **Review Depth:** Light (small, isolated regex fix)

## PR Description Assessment
Clear and sufficient — explains the root cause (Chrome 151+ prefixing the target URL with `--url`), the fix, and test results. Matches the code.

## Existing Feedback Status
None — first review. (A follow-up commit landed mid-review adding one more test case; reviewed against the latest commit, `b3e50fc`.)

---

## Findings

No MUST FIX / SHOULD FIX issues found.

The regex change is minimal and correct:
```ts
const urlMatch = curlInput.match(
  /(?:curl\s+|--url(?:\s+|=))['"]?(https?:\/\/([\w.-]+)\.slack\.com[^'"\s]*)/
);
```
- Correctly handles both `--url 'https://...'` (space) and `--url="https://..."` (equals) forms, plus the existing positional form.
- Preserves enterprise-domain extraction (`chromeorg.enterprise.slack.com` → workspace `chromeorg`), matching the existing `extractSlackWorkspaceName` split-on-first-`.` behavior.
- No catastrophic-backtracking risk — bounded alternation, no nested quantifiers.
- The added "decoy header" test (`referer: https://decoyteam.slack.com/` before the real `--url`) confirms the pattern doesn't false-positive on a Slack URL that merely appears elsewhere in the command — good defensive coverage for exactly the kind of edge case worth checking on an unanchored regex like this.

Test coverage is solid: both new `--url` spellings are tested individually, in the shared "real-world formats" round-trip loop, and against the decoy-header case. CI (unit, integration, CodeQL) is green.

---

## Summary

| Category | Must Fix | Should Fix | Consider | Questions |
|----------|----------|------------|----------|-----------|
| Correctness | | | | |
| Code Quality | | | | |
| Security | | | | |
| Performance | | | | |
| Architecture | | | | |
| UI/UX | | | | |
| Cross-Platform | | | | |
| Observability | | | | |
| Testing | | | | |
| Documentation | | | | |

## Documentation Needs
- [x] No documentation changes needed

## Verdict

**APPROVE**

Small, focused, well-tested fix matching the linked issue's (`#94`, `ready-for-pr`) suggested implementation. No correctness, security, or performance concerns; CI is green.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #98, opened 2026-08-06
ci: warn on unsigned commits in pull requests

### LINKED ISSUE #92: ci: warn on unsigned commits in PRs so contributors find out before review, not at merge
## What

A GitHub Actions workflow that runs on every pull request, checks whether all commits in the PR are signed and verified, and — when they are not — posts a single comment explaining how to sign them. The comment must be updated in place on later runs rather than duplicated, and resolved/removed once the commits are verified.

## Why

`main` is protected by a ruleset (`default-branch-baseline`) that requires signed commits, and the ruleset has no bypass actors — so no maintainer, including an admin, can merge a PR with unsigned commits.

Today nothing surfaces this until merge time. A contributor gets through review, approval, and a full green CI run, and only then discovers the PR is unmergeable for a reason that has nothing to do with their code. The fix is a force-push, which invalidates the review that was already given.

This just happened on #91: approved, all checks passing, and blocked on 7 unsigned commits. The reviewer had to diagnose the ruleset by hand and write the remediation comment manually. That is work that will repeat for every external contributor, and it is entirely mechanical.

Catching it on the first push moves the cost to the cheapest possible moment — before review, when a rebase is free.

## How (a suggestion)

- Trigger on `pull_request` (`opened`, `synchronize`, `reopened`).
- Read commit verification via `GET /repos/{owner}/{repo}/pulls/{number}/commits` and inspect `commit.verification.verified`. Prefer this over shelling out to `git verify-commit`, which does not know about GitHub-hosted GPG/SSH keys or the web-flow signer.
- **Deduplication:** find an existing bot comment by a hidden marker (e.g. `<!-- signed-commits-check -->`) in the body, then `PATCH` it instead of creating a new one. Listing comments and matching on prose is brittle.
- When every commit verifies, edit the comment to a short "all commits verified" note (or delete it) so a stale warning never outlives the problem.
- List the specific unsigned SHAs — on a long branch, "some commits are unsigned" is not actionable.
- Include the remediation inline, both the rebase and the SSH-signing setup:
  ```
  git rebase --exec 'git commit --amend --no-edit -S' origin/main
  git push --force-with-lease
  ```
- Permissions: `pull-requests: write`, `contents: read`. Use the default `GITHUB_TOKEN`.
- Decide whether this is advisory (comment only) or a required check. Advisory is the safer start: the ruleset already blocks the merge, so a failing check adds noise without adding protection.

### Worth considering

Forks. `pull_request` gives a read-only token for fork PRs, which is exactly the case that needs the comment most. `pull_request_target` has write access but must not check out PR code — safe here, since this workflow only needs the API and never runs contributor code.

vibexp-gh:
Closes #92

## What

Adds `.github/workflows/signed-commits.yml`, an **advisory** workflow that runs on every PR push, reads commit verification from the GitHub API, and maintains a **single self-updating comment** listing any unsigned commits and how to fix them. Also documents the signing requirement in `CONTRIBUTING.md`.

## Why

`main` is protected by the `default-branch-baseline` ruleset, which requires signed commits and has **no bypass actors** — one unsigned commit makes a PR unmergeable for everyone, maintainers included. Today that only surfaces at merge time, after review and a full green CI run, and the fix is a force-push that invalidates the approval already given (exactly what happened on #91). This moves the discovery to the first push, when a rebase is free.

## How

- **Trigger:** `pull_request_target` on `opened`/`synchronize`/`reopened`. Deliberate over `pull_request`: fork PRs are the case that most needs this comment, and `pull_request` hands forks a read-only token. Safe here because the job **never checks out or executes PR code** — it only reads the API.
- **Source of truth:** `GET /pulls/{n}/commits` → `commit.verification.verified`, paginated. Not `git verify-commit`, which doesn't know about GitHub-hosted keys or the web-flow signer.
- **Dedup:** finds its own comment by the hidden marker `<!-- signed-commits-check -->` and `PATCH`es it. Idempotent — if the body is unchanged it makes no API call at all, so a series of pushes doesn't churn the timeline.
- **Self-resolving:** once every commit verifies, the comment is rewritten to a short "all verified" note. A stale warning never outlives the problem. If all commits are verified and no comment exists, the job does nothing.
- **Advisory, never red.** The ruleset already blocks the merge; a failing check would add noise without adding protection.
- **Permissions:** `contents: read`, `pull-requests: write`, default `GITHUB_TOKEN`. Action pinned by SHA, matching the convention in the other workflows.
- **Concurrency group per PR** with `cancel-in-progress`, so rapid pushes can't race on the comment.

### Hardening

- Commit subjects are contributor-controlled and land inside a markdown table, so they're clipped to 72 chars and `` | ` < > \ `` are escaped. Without this, a crafted commit subject breaks the table or injects markup into a comment written by a token with `pull-requests: write`.
- The comment lookup requires `user.type === 'Bot'`. A human can paste the marker into their own comment, and the default token **cannot** edit someone else's — that would hard-fail the job on a PR that is otherwise fine.
- Handles a missing `verification` object (reports `unknown`) and notes when the 250-commit API cap truncates the list.

## Testing

Workflow logic can't run in CI before it's on `main`, so the embedded script was extracted and exercised against a mocked Octokit across six cases, all passing:

| Case | Expected | Result |
|---|---|---|
| Unsigned commits, no existing comment | create comment | ✅ |
| Unsigned, existing identical comment | no API call (idempotent) | ✅ |
| All verified, existing warning comment | patched to "all verified" | ✅ |
| All verified, no comment | no-op | ✅ |
| Commit with no `verification` field | listed, reason `unknown` | ✅ |
| Hostile subject + human comment carrying marker | subject escaped; creates own comment instead of 403-ing | ✅ |

YAML parses; the embedded script passes `node --check` as an async body.

## Note on rollout

`pull_request_target` workflows run from the **base** branch, so this will not run against this PR — it takes effect on the first PR opened after merge.

### DIFF
--- a/.github/workflows/signed-commits.yml
+++ b/.github/workflows/signed-commits.yml
@@ -16,6 +16,10 @@ name: Signed Commits
 
 on:
   pull_request_target:
+    # Scoped to the protected branches, matching ci.yml and test.yml. The
+    # comment asserts that the base branch requires signed commits, so it must
+    # not fire for a base branch where that is untrue.
+    branches: [ main, master ]
     types: [opened, synchronize, reopened]
 
 # Least-privilege default token for every job in this workflow.
@@ -142,14 +146,16 @@ jobs:
 
             const body = unsigned.length > 0 ? warning : resolved;
 
+            // GitHub may hand comment bodies back with CRLF line endings, so a
+            // raw === would report "changed" on every push and re-edit an
+            // identical comment forever.
+            const same = (a, b) => a.replace(/\r\n/g, '\n') === b.replace(/\r\n/g, '\n');
+
             if (unsigned.length === 0 && !existing) {
               // Nothing to say and nothing stale to clean up.
               core.info('All commits verified; no existing comment to update.');
-              return;
-            }
-
-            if (existing) {
-              if (existing.body === body) {
+            } else if (existing) {
+              if (same(existing.body, body)) {
                 core.info('Existing comment is already up to date.');
               } else {
                 await github.rest.issues.updateComment({
--- a/.github/workflows/signed-commits.yml
+++ b/.github/workflows/signed-commits.yml
@@ -1,0 +1,180 @@
+name: Signed Commits
+
+# Advisory check: tells contributors as early as possible when a PR contains
+# commits GitHub cannot verify. `main` is protected by a ruleset that requires
+# signed commits and has no bypass actors, so an unsigned commit makes the PR
+# unmergeable — for anyone, including admins. Surfacing it on the first push
+# means the fix (a rebase) is free, instead of invalidating an approval later.
+#
+# This job never fails the PR. The ruleset already blocks the merge; a red
+# check would add noise without adding protection.
+#
+# `pull_request_target` (not `pull_request`) is deliberate: fork PRs are the
+# case that most needs this comment, and `pull_request` hands forks a
+# read-only token. This is safe here because the job never checks out or
+# executes PR code — it only reads the GitHub API.
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+
+# Least-privilege default token for every job in this workflow.
+permissions:
+  contents: read
+  pull-requests: write
+
+# One run per PR; a rapid series of pushes should not race on the comment.
+concurrency:
+  group: signed-commits-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  check-signatures:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check commit signatures and report
+        uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
+        with:
+          script: |
+            const MARKER = '<!-- signed-commits-check -->';
+            const { owner, repo } = context.repo;
+            const pr = context.payload.pull_request;
+            const baseRef = pr.base.ref;
+
+            // The PR commits endpoint is capped at 250 commits by GitHub.
+            const commits = await github.paginate(github.rest.pulls.listCommits, {
+              owner,
+              repo,
+              pull_number: pr.number,
+              per_page: 100,
+            });
+            const truncated = commits.length >= 250;
+
+            const unsigned = commits.filter((c) => !c.commit?.verification?.verified);
+            core.info(`${commits.length - unsigned.length}/${commits.length} commits verified`);
+
+            // Only ever adopt a comment this workflow could have written. A
+            // human can paste the marker into their own comment, and the
+            // default token cannot edit someone else's — that would hard-fail
+            // the job on a PR that may be perfectly fine.
+            const existing = (
+              await github.paginate(github.rest.issues.listComments, {
+                owner,
+                repo,
+                issue_number: pr.number,
+                per_page: 100,
+              })
+            ).find((c) => c.user?.type === 'Bot' && c.body?.includes(MARKER));
+
+            // Commit subjects are contributor-controlled and land inside a
+            // markdown table, so neutralise anything that could break out of
+            // the cell or inject markup.
+            const cell = (text) => {
+              const raw = text.replace(/\r/g, '').trim();
+              const clipped = raw.length > 72 ? `${raw.slice(0, 71)}…` : raw;
+              return clipped.replace(/[|`<>\\]/g, '\\$&') || '(no subject)';
+            };
+
+            const list = unsigned
+              .map((c) => {
+                const subject = cell(c.commit.message.split('\n')[0]);
+                const reason = c.commit.verification?.reason ?? 'unknown';
+                return `| \`${c.sha.slice(0, 7)}\` | ${subject} | \`${reason}\` |`;
+              })
+              .join('\n');
+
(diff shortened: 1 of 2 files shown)


shahariaazam (approved):
## PR Overview

- **Title:** ci: warn on unsigned commits in pull requests
- **Scope:** 2 files changed, +208/-0 lines, 2 commits
- **Category:** CI / Config
- **Risk Level:** Low-Medium (adds a `pull_request_target` workflow with `pull-requests: write`)
- **Review Depth:** Standard (small diff, but elevated-permission trigger warrants a security-focused read)

## PR Description Assessment
Clear and thorough — explains the trigger choice (`pull_request_target` vs `pull_request`), why it's safe (no checkout/execution of PR code), the dedup/self-resolving comment strategy, and the hardening applied to contributor-controlled commit subjects. Code matches the description.

## Existing Feedback Status
Second commit (`431cad0f`) is a self-addressed review pass: scopes the trigger to `main`/`master` (matching `ci.yml`/`test.yml`, and correctly avoiding the workflow asserting "this branch requires signed commits" on a branch where that isn't true), normalizes CRLF before the idempotency comparison (fixing a real bug where GitHub-returned CRLF bodies would have defeated the dedup and re-edited an identical comment on every push), and writes the job summary on the all-verified path too. All three are correct fixes; nothing outstanding from that pass.

---

## Findings

### SHOULD FIX — "Bot" comment adoption is broader than the stated safety net

- **Location:** `.github/workflows/signed-commits.yml:56-64`
- **Category:** Security / Correctness
- **Code:**
  ```js
  const existing = (
    await github.paginate(github.rest.issues.listComments, { ... })
  ).find((c) => c.user?.type === 'Bot' && c.body?.includes(MARKER));
  ```
- **Issue:** The comment says "the default token cannot edit someone else's [comment] — that would hard-fail the job." That's not a documented guarantee — GitHub's REST API generally authorizes comment updates by repository permission level, not by resource ownership, for tokens with sufficient write access. If that assumption doesn't hold, `pull-requests: write` is enough to edit *any* comment, not just ones this workflow created. Since the match is only `type === 'Bot'` (not this workflow's specific identity), any other GitHub App/bot commenting on the PR whose body happens to contain `<!-- signed-commits-check -->` would get silently overwritten with this workflow's warning/resolved text. Low likelihood, but the blast radius ("clobber another bot's comment on a fork PR") is avoidable for free.
- **Suggestion:**
  ```js
  ).find((c) => c.user?.login === 'github-actions[bot]' && c.body?.includes(MARKER));
  ```
  Strictly safer regardless of whether the assumed 403 protection actually exists, and makes the ownership check explicit rather than inferred from `type`.

### CONSIDER — Ship the check script as a tested file, not only inline YAML

- **Location:** `.github/workflows/signed-commits.yml:37-182`
- **Category:** Testing
- **Issue:** The PR body describes 6 test cases run against a mocked Octokit, which is good rigor — but that test suite isn't part of this PR, so there's no regression protection for this script in the repo going forward, and (per the PR's own "Note on rollout") the workflow can't even run against this PR since `pull_request_target` executes from `main`. The next person touching this logic has no safety net, and the fix in the second commit (CRLF handling) is exactly the kind of thing a committed test would have caught up front.
- **Suggestion:** Extract the script into `.github/scripts/check-signed-commits.js`, have the workflow load it, and commit the 6 mocked-Octokit cases as a `bun test` file alongside it. Reasonable as a fast follow rather than blocking this PR.

### Verified — no issue
- Pinned SHA `actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8` checked against the upstream `v8.0.0` release tag — matches. Supply-chain pin is correct.
- `pull_request_target` usage is safe here: no `actions/checkout` step, no execution of PR-supplied code — only GitHub API reads/writes via `github-script`. Avoids the classic "pwn request" pattern.
- Table-cell escaping (`cell()`) correctly neutralizes `|`, `` ` ``, `<`, `>`, `\` in contributor-controlled commit subjects before they're embedded in the markdown table.
- Self-resolving/idempotent comment logic is correct across all traced states (create / update-changed / no-op-unchanged / no-op-nothing-to-say), including after the CRLF fix in the second commit.
- CONTRIBUTING.md addition is accurate and consistent with the workflow's actual behavior.

---

## Summary

| Category | Must Fix | Should Fix | Consider | Questions |
|----------|----------|------------|----------|-----------|
| Correctness | | | | |
| Code Quality | | | | |
| Security | | 1 | | |
| Performance | | | | |
| Architecture | | | | |
| UI/UX | | | | |
| Cross-Platform | | | | |
| Observability | | | | |
| Testing | | | 1 | |
| Documentation | | | | |

## Documentation Needs
- [x] No documentation changes needed (CONTRIBUTING.md update in this PR already covers it)

## Verdict

**APPROVE**

Well-designed, appropriately cautious use of `pull_request_target` (no code checkout/execution) with solid input-sanitization for contributor-controlled commit subjects, and a correct self-resolving/idempotent comment lifecycle after the second commit's CRLF fix. One low-severity hardening suggestion (narrow the "existing comment" match to the actual bot identity) and one testing-debt note (ship the mocked test cases in-repo) — neither blocks merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #119, opened 2026-08-20
security: verify the downloaded binary before installing it (#104)

### LINKED ISSUE #104: self-update installs the downloaded binary without verifying its integrity
Found while auditing v0.8.0 before rolling it out on a work machine, where the CLI holds live corporate Slack credentials. Filing per SECURITY.md, which asks for a public issue unless the finding is a zero-day or actively exploited — this is neither, it is hardening of the update path. Pre-existing: the code is byte-identical in v0.7.0, so it is not a v0.8.0 regression.

## WHAT

`slackcli update` downloads the release asset and installs it over the running binary without checking that the bytes are the ones the release published.

`src/lib/updater.ts` (v0.8.0), around lines 139-175:

```ts
const asset = release.assets.find(a => a.name === binaryName);
...
const response = await fetch(asset.browser_download_url);
...
const buffer = await response.arrayBuffer();
const tmpPath = join(tmpdir(), `slackcli-update-${Date.now()}`);

await writeFile(tmpPath, new Uint8Array(buffer));
await chmod(tmpPath, 0o755);
...
await rename(tmpPath, currentBinary);
```

Two things:

1. **No integrity check.** Nothing compares the downloaded bytes against a known digest, even though the release publishes `checksums.txt` and the GitHub API already returns a per-asset digest (verified against v0.8.0):

   ```
   $ gh api repos/shaharia-lab/slackcli/releases/tags/v0.8.0 \
       -q ".assets[] | select(.name|test(\"macos-arm64\")) | .digest"
   sha256:1104f5fd5957bd12953ad67b9a04b4748d64c8474faa1be010de2050c5391108

   $ shasum -a 256 slackcli-macos-arm64
   1104f5fd5957bd12953ad67b9a04b4748d64c8474faa1be010de2050c5391108
   ```

   The material to verify against is already in the response the updater parses — `GitHubRelease.assets` just does not declare the field.

2. **Predictable path in a shared temp directory.** `slackcli-update-${Date.now()}` is guessable and `writeFile` does not use `O_EXCL`, so on a multi-user host another user can pre-create or race that path. The file is then `chmod 0o755` and renamed over `process.execPath`.

## WHY

The updater is the one code path that writes an executable and installs it as the CLI itself, and this CLI holds Slack tokens. Anything that manages to substitute those bytes — a compromised or mistakenly re-uploaded release asset, a TLS-terminating proxy on a corporate network, a same-host user winning the temp-path race — gets code execution as the user, with the token store one file away.

The check is also close to free: the digest is already in the API response, so it costs one hash of a buffer that is already in memory. Right now `checksums.txt` is published every release and never used by anything.

Note the practical effect today: because the updater does not verify, someone being careful has to bypass `slackcli update` entirely and install by hand (download the asset, compare against `checksums.txt`, replace the binary) — which is what I ended up doing. Verifying in the updater removes that reason to work around it.

## HOW (suggestion)

Small and self-contained, no new dependencies:

1. Declare the field the API already sends:

   ```ts
   assets: Array<{
     name: string;
     browser_download_url: string;
     digest?: string; // "sha256:<hex>", present on current GitHub releases
   }>;
   ```

2. After `arrayBuffer()`, before anything touches the disk:

   ```ts
   if (asset.digest) {
     const actual = `sha256:${createHash("sha256").update(new Uint8Array(buffer)).digest("hex")}`;
     if (actual !== asset.digest) {
       throw new Error(`Checksum mismatch for ${binaryName}: expected ${asset.digest}, got ${actual}`);
     }
   }
   ```

   Guarded on presence so an older release without the field still updates. If you would rather fail closed, treating a missing digest as an error is also reasonable — every current release has it.

3. Replace the predictable temp path with a private directory:

   ```ts
   const tmpDir = await mkdtemp(join(tmpdir(), "slackcli-update-"));
   const tmpPath = join(tmpDir, binaryName);
   ```

   `mkdtemp` creates with 0700 and fails rather than reusing an existing path, which closes the race; remove the directory in the existing failure path.

An alternative to (1)/(2) is downloading `checksums.txt` from the same release and matching the line for the asset. The `digest` field is fewer moving parts, but the checksums file is the thing you already publish, so either is fine — your call which fits better.

Happy to open the PR once this is triaged and labelled `ready-for-pr`, per the Repository Constitution in CLAUDE.md. Tests for the mismatch path would go alongside; tell me if you would prefer the `checksums.txt` variant instead.

---
Reported with the help of an AI agent (Claude); the code paths, the API response and the checksum comparison above were each verified by hand against v0.8.0 before filing.

victor-develop:
Closes #104. Following your call in the issue: `digest` field rather than `checksums.txt`, fail closed on a missing digest, `mkdtemp` in the same PR, backup-restore in the catch block untouched.

## What changed

`src/lib/updater.ts`:

- `GitHubRelease.assets` now declares `digest?: string` — the `"sha256:<hex>"` GitHub already sends.
- `verifyAssetDigest(binaryName, bytes, expected)` hashes the buffer and compares, called right after `arrayBuffer()` and before anything touches the disk.
- A missing digest, or one whose algorithm is not `sha256`, throws. The error tells the user how to install by hand instead.
- The staging path is now `mkdtemp(join(tmpdir(), 'slackcli-update-'))` + the binary name inside it. `mkdtemp` creates with 0700 and fails rather than reusing a path, so the pre-created-file race on the old `slackcli-update-${Date.now()}` is gone. The directory is removed in a `finally`, on both paths.
- The write/chmod and the install now share one `try` so the cleanup covers a failed write too. The `catch` still logs and rethrows, with the backup-restore TODO left as it was.

Hex comparison is case-insensitive: GitHub sends lowercase, but a digest that only differs in case is the same digest, and rejecting it would be a false alarm about a tampered download.

## Tests

`src/lib/updater.test.ts`, 9 new cases:

- `verifyAssetDigest` — match, uppercase hex, mismatch, no digest, `md5:` prefix.
- `performUpdate` end-to-end against a stubbed `fetch` and a throwaway `process.execPath`: mismatched bytes and a digest-less asset both reject, and in each case the file that was already installed still holds its original contents. The matching-digest case installs the new bytes and leaves no `slackcli-update-*` directory behind.

`bun test` 412 pass / 0 fail, `bun run type-check` clean. `pre-commit` is not installed on this machine, so those hooks ran only in the form CI runs them (type-check + tests).

## Not covered

`rename()` from `tmpdir()` to `process.execPath` still fails with `EXDEV` when they are on different filesystems — pre-existing, unchanged by this PR, and out of scope for the integrity fix.

### DIFF
--- a/src/lib/updater.test.ts
+++ b/src/lib/updater.test.ts
@@ -1,5 +1,9 @@
-import { describe, expect, it } from 'bun:test';
-import { isNewerVersion, isInstalledViaHomebrew, getUpdateCommand, getCurrentVersion, performUpdate } from './updater.ts';
+import { afterEach, describe, expect, it } from 'bun:test';
+import { isNewerVersion, isInstalledViaHomebrew, getUpdateCommand, getCurrentVersion, performUpdate, verifyAssetDigest } from './updater.ts';
+import { createHash } from 'crypto';
+import { mkdtemp, readFile, readdir, rm, writeFile } from 'fs/promises';
+import { tmpdir } from 'os';
+import { join } from 'path';
 import packageJson from '../../package.json';
 
 describe('isNewerVersion', () => {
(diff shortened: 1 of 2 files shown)


vibexp-agent (approved):
Thanks for this — a genuine security fix, tightly scoped, and you followed the issue-first process end to end including waiting for `ready-for-pr`.
Verified locally and in CI: type-check clean, 412 tests pass, and `digest` is confirmed present on unauthenticated release responses so failing closed is safe.

PR #120, opened 2026-08-20
fix(messages): send parse=none on chat.update so links survive an edit (#106)

### LINKED ISSUE #106: messages edit breaks every link in the message it edits (chat.update missing parse=none)
## WHAT

`slackcli messages edit` stores the new text HTML-escaped, so any Slack link markup in it (`<url|label>`) is saved as `&lt;url|label&gt;` and renders as literal text. The message updates fine; the links in it die. `slackcli messages send` with the exact same string is unaffected.

Reproduced on 0.8.0, browser (xoxc/xoxd) auth:

```bash
slackcli messages send --recipient-id C123 --message 'A <https://example.com|label> B'
# stored text: A <https://example.com|label> B          -> blocks contain a `link` element

slackcli messages edit --channel-id C123 --timestamp <ts> --message 'A <https://example.com|label> B'
# stored text: A &lt;https://example.com|label&gt; B    -> renders as literal text
```

## WHY

Any workflow that maintains a message in place — a status message, a checklist, a digest that gets refreshed — loses all of its links the first time it is edited. The failure is silent: the API call succeeds, and the damage is only visible in the rendered message.

It also makes `send` and `edit` inconsistent for identical input, which is surprising: users reasonably expect the same text to produce the same message.

## HOW (suggested)

The escaping is Slack's, not slackcli's: `chat.update` treats the text the way the client would unless `parse=none` is passed, while `chat.postMessage` already defaults to `parse=none`. Passing it explicitly on update restores the symmetry.

`src/lib/slack-client.ts`:

```ts
async updateMessage(channel: string, ts: string, text: string): Promise<any> {
  return this.request('chat.update', { channel, ts, text, parse: 'none' });
}
```

Verified against a real workspace — a parameter matrix on `chat.update` with a browser token:

| params | `<url\|label>` survives |
|---|---|
| (none) | no |
| `parse=none` | **yes** |
| `parse=full` | no |
| `mrkdwn=true` | no |
| `parse=none` + `mrkdwn=true` | yes |
| `link_names=false` | no |

I have the fix and a test for it ready on a branch. Per CONTRIBUTING.md I am waiting for this issue to be triaged and labelled `ready-for-pr` before opening the PR — happy to send it the moment it is.

victor-develop:
Closes #106, which you labelled `ready-for-pr`.

## What changed

`src/lib/slack-client.ts` — one parameter:

```ts
return this.request('chat.update', { channel, ts, text, parse: 'none' });
```

The comment above it records why it has to be explicit, since the value looks redundant next to `chat.postMessage`: [the docs](https://docs.slack.dev/reference/methods/chat.update) say `parse` "Defaults to `client`, unlike `chat.postMessage`", and also that "If you do not specify a value for this field, the original value set for the message will be overwritten with the default" — so omitting it does not inherit what the message was posted with.

## Verified on a real workspace

v0.9.0 from source against browser (xoxc/xoxd) auth, in a DM, same text every time, reading `text` back with `conversations.history`:

| step | stored `text` |
|---|---|
| posted with `chat.postMessage` | `<https://example.com\|labelled link>` |
| edited by v0.9.0 (no `parse`) | `&lt;https://example.com\|labelled link&gt;` |
| edited with `parse: 'none'` | `<https://example.com\|labelled link>` |

Two details from the repro worth having on record:

- The corruption does not compound. Editing three more times leaves it at `&lt;…&gt;` rather than escaping the `&`, so a message is broken once and stays that way. It does not self-heal either — text posted as a link never becomes a link again without a corrected edit.
- `<@U…>` mentions came through all three steps intact, so this is specifically the `<url|label>` markup, not user references. `link_names` also defaults to `none` on `chat.update` per the docs; I left it alone since nothing in this repro needed it.

## Tests

`src/lib/slack-client.test.ts`: the existing `updateMessage` assertion now includes `parse: 'none'` in the expected params, plus one case named for the failure it prevents. The test double records `{method, params}`, so it pins the wire call — which is where the bug was.

`bun test` 405 pass / 0 fail, `bun run type-check` clean. `pre-commit` is not installed on this machine, so those hooks ran only in the form CI runs them.

## Not covered

`chat.postMessage` is untouched: its `parse` default is already `none`, and the repro confirms posted text arrives unescaped.

### DIFF
--- a/src/lib/slack-client.test.ts
+++ b/src/lib/slack-client.test.ts
@@ -127,9 +127,25 @@ describe('SlackClient.updateMessage', () => {
     expect(client.calls).toEqual([
       {
         method: 'chat.update',
-        params: { channel: 'C123', ts: '1234567890.123456', text: 'Corrected message' },
+        params: {
+          channel: 'C123',
+          ts: '1234567890.123456',
+          text: 'Corrected message',
+          parse: 'none',
+        },
       },
     ]);
     expect(response.ts).toBe('1234567890.123456');
+  });
+
+  // Without this, Slack applies the chat.update default (`client`) and stores
+  // `&lt;https://example.com|label&gt;`, which renders as literal text: every
+  // link in an edited message dies, silently, on a call that returns ok.
+  it('sends parse=none so link markup survives the edit', async () => {
+    const client = new TestSlackClient();
+
+    await client.updateMessage('C123', '1234567890.123456', 'A <https://example.com|label> B');
+
+    expect(client.calls[0]!.params.parse).toBe('none');
   });
 });
--- a/src/lib/slack-client.ts
+++ b/src/lib/slack-client.ts
@@ -153,7 +153,10 @@ export class SlackClient {
   }
 
   async updateMessage(channel: string, ts: string, text: string): Promise<any> {
-    return this.request('chat.update', { channel, ts, text });
+    // `parse` must be explicit: chat.update defaults it to `client` (unlike
+    // chat.postMessage, which defaults to `none`), and that escapes the
+    // `<url|label>` markup in the text, so editing a message kills its links.
+    return this.request('chat.update', { channel, ts, text, parse: 'none' });
   }
 
   async uploadFileExternal(channel: string, filePath: string, options: {


vibexp-agent (approved):
Thanks for the fix. Sending `parse: 'none'` on `chat.update` stops Slack applying its `client` default and escaping `<url|label>`, so links survive an edit.

PR #124, opened 2026-08-21
feat(messages): support native Block Kit tables and Markdown

### LINKED ISSUE #121: feat: support Block Kit JSON in messages send
## What

Add an optional structured-blocks input to `slackcli messages send` so callers can pass a JSON array of [Slack Block Kit blocks](https://docs.slack.dev/reference/block-kit/blocks/) to `chat.postMessage`, including:

- Native [`table` blocks](https://docs.slack.dev/reference/block-kit/blocks/table-block/) with column settings and rich-text cells.
- Native [`markdown` blocks](https://docs.slack.dev/reference/block-kit/blocks/markdown-block/) that let Slack render standard Markdown, including headings, links, lists, task lists, syntax-highlighted code blocks, and Markdown tables.

A compact CLI shape could be:

```bash
# Load any Block Kit array from a file
slackcli messages send \
  --recipient-id=C1234567890 \
  --message="Project status" \
  --blocks=@blocks.json

# Or pass a native markdown block inline
slackcli messages send \
  --recipient-id=C1234567890 \
  --message="Release notes" \
  --blocks='[{"type":"markdown","text":"# Release notes\n\n- [x] Build\n- [ ] Deploy\n\nSee the [runbook](https://example.com/runbook)."}]'
```

Inline JSON should also be supported. Existing text and file sends should remain compatible, and the structured-block path should work with both standard Slack tokens and browser-session authentication.

## Why

The current command only accepts message text and file attachments. Users who need richer structured output must flatten it into the top-level message text or a Markdown/CSV-like code block.

Slack's native table block supports columns, wrapping, alignment, formatted headers, and rich-text cells such as links. Slack's native markdown block accepts standard Markdown and leaves translation/rendering to Slack, which is useful for content such as LLM responses that can lose formatting when converted manually. Exposing general Block Kit JSON is both simpler and more reusable than adding table-specific or markdown-specific commands.

## Suggested approach

- Add `messages send --blocks <json|@file>` for an inline JSON array or an `@`-prefixed JSON file.
- Validate that the input parses to an array of block objects with a `type`.
- Keep `--message` as the notification and accessibility fallback text.
- Pass the array to `chat.postMessage` for standard-token auth.
- JSON-encode structured values when browser-session auth submits URL-encoded form data.
- Document copyable native table and native markdown examples.

## Acceptance criteria

- A documented command sends a top-level native `table` block through `chat.postMessage`.
- A documented command sends a top-level native `markdown` block through `chat.postMessage` without converting it to Slack mrkdwn or rich text first.
- Rich-text links in table cells and standard Markdown links in markdown blocks are preserved in the request payload.
- Existing text and file send behavior remains compatible.
- Standard-token and browser-session authentication are covered.
- Parsing/validation and request-shape tests include both table and markdown blocks.

lelouvincx:
## Summary

- add `messages send --blocks <json|@file>` for inline or file-based Block Kit arrays
- validate the top-level array and block `type` fields before sending
- pass native `table`, `markdown`, and other Block Kit blocks unchanged to `chat.postMessage`
- JSON-encode structured request values for browser-session authentication
- preserve existing text, thread reply, DM, and file-upload behavior
- document copyable native table and native Markdown examples

## Tests

- `bun test` — 421 passed, 0 failed
- `bun run type-check`
- `bun run build`
- `pre-commit run --all-files`

Closes #121


src/commands/messages.ts
```diff
@@ -63,8 +104,10 @@ export function createMessagesCommand(): Command {
           return;
         }
 
+        const blocks = options.blocks ? await parseBlocksInput(options.blocks) : undefined;
```
vibexp-agent, line 110: This sits after the `--file` early return, so `--blocks` is silently dropped when both flags are passed: the file uploads and the command still reports success. Worth either erroring on the combination or passing the blocks through to `uploadFileExternal`.

vibexp-agent (commented):
Adds `messages send --blocks` for inline or @file Block Kit JSON, and JSON-encodes object params in the browser-auth form body so both auth paths carry blocks. Verified on top of main: type-check clean, 421 tests pass, and reverting the browserRequest change does make the new browser-encoding test fail, so it is a real guard. Confirmed against the Slack docs that `table` and `markdown` are supported blocks for chat.postMessage.

One note inline: `--blocks` is silently ignored when `--file` is also passed. Thanks for the contribution.

vibexp-agent (changes requested):
Requesting changes for the inline note: `--blocks` is silently dropped when `--file` is also passed.

Everything else verified clean on top of main: type-check, 421 tests, and the new browser-encoding test does fail when the change is reverted.

src/commands/messages.ts
lelouvincx: Fixed in https://github.com/shaharia-lab/slackcli/pull/124/commits/4b6801265645c33b31967a5b882096dc9bbc6648. `--blocks` now conflicts with `--file` during Commander option parsing, so the command exits before authentication or upload. I also added a regression test and updated the README. The full local suite passes with 422 tests.

vibexp-agent (commented):
The --blocks/--file conflict is fixed, and the new test does fail when the conflicts('file') is removed, so it is a real guard. Nothing further from me on this head.

Thanks for adding native Block Kit table and markdown support to `messages send`, on both auth paths.

vibexp-agent (approved):

PR #149, opened 2026-09-01
feat(emoji): add read-only emoji list and emoji get commands

### LINKED ISSUE #140: Add a read-only `emoji` command group (list + get workspace custom emoji)
## What

Add a read-only `emoji` command group with two subcommands:

- `emoji list` — list the workspace's custom emoji, sorted by name. Options: `--limit`, `--no-aliases` (originals only), `--workspace`, `--json`.
- `emoji get <name>` — show one emoji's details (original + image URL, or alias + its target). Accepts the name with or without surrounding colons.

## Why

`slackcli` can read messages, conversations, canvases, saved items, and search results, but there's no way to inspect a workspace's custom emoji from the CLI. Custom emoji are workspace metadata that's useful to audit and script against (e.g. exporting the set, checking whether a name exists before referencing it, resolving what an alias points at). It fills a read gap consistent with the existing read command groups.

## How (suggested)

Follows the repo's documented add-a-command pattern:

- `SlackClient.listEmoji()` → Slack's `emoji.list`, routed through `this.request()` so it works for both standard and browser auth.
- A `src/lib/emoji.ts` module normalizes Slack's flat `{name: value}` map into a typed list, distinguishing originals (image URL) from aliases (`alias:<target>`).
- A formatter for terminal output + `--json`, a `commands/emoji.ts` group wired into `src/index.ts`, tests, and a `docs/user-guide/emoji.md` page.

Consistent with the checklist conventions (`--json`, `--workspace`, no `process.exit()` after `writeJson()`, `ora` spinner). Implementation is already prototyped and passing type-check + tests locally; a PR would follow once this issue is triaged with `ready-for-pr`.

tmack8001:
## What

Adds a read-only `emoji` command group for viewing a workspace's custom emoji, working under both standard (`xoxb`/`xoxp`) and browser (`xoxc`/`xoxd`) auth via `emoji.list`.

- `emoji list` — name-sorted list, with `--limit`, `--no-aliases`, `--workspace`, and `--json`
- `emoji get <name>` — details for a single emoji, accepting the name with or without surrounding colons

## Why

Custom emoji are a gap in the read commands (as noted in #140). This fills it with a read-only surface that mirrors the existing `--json` / `--workspace` conventions and the `slack-client` auth abstraction, so it works the same across both auth types.

## How

- `src/lib/emoji.ts` — `normalizeEmojiMap` turns the flat `emoji.list` map into a typed, name-sorted list, distinguishing originals (`url`) from aliases (`is_alias` + `alias_for`); `fetchCustomEmoji` / `getCustomEmoji` wrap the fetch and single-lookup.
- `src/lib/slack-client.ts` — `listEmoji()` routes through the shared `request()` path so both auth types work.
- `src/commands/emoji.ts` — the `emoji list` / `emoji get` group, following the `saved.ts` / `docs/development/adding-a-command.md` pattern.
- `src/lib/formatter.ts` — `formatEmojiList` / `formatEmoji` in the existing chalk style.
- `src/types/index.ts` — `CustomEmoji`, `EmojiListOptions`, `EmojiGetOptions`.
- Docs + tests: `docs/user-guide/emoji.md`, `src/lib/emoji.test.ts` + `src/commands/emoji.test.ts` (18 tests total), and `CLAUDE.md` architecture updates.

## Design notes (from #140 discussion)

- **`--limit` is client-side truncation, applied after `--no-aliases`.** `emoji.list` returns one flat map with no cursor, so the command fetches the whole set, filters aliases (when `--no-aliases`), sorts by name, then slices by `--limit`. `--json` respects it: `emoji_count` reflects the truncated set, matching `item_count` / `canvas_count` elsewhere.
- **`emoji get` resolves one hop.** For an alias it reports `is_alias: true` and `alias_for: <immediate target>`, not the full chain to a terminal image — the default you confirmed in #140.
- **No `process.exit()` after `writeJson()`** (per #73): both subcommands `return` after writing JSON; `process.exit(1)` only occurs on error / not-found paths, before any JSON is written.

## Tests

- `src/lib/emoji.test.ts` (15 tests) covers map normalization, alias detection, name sorting, single-`get` name matching (with and without colons), and `parseEmojiLimit` — including the regression cases where a non-numeric (`--limit abc`), negative (`--limit -2`), or zero limit is rejected with a clear error instead of `slice(0, NaN)` silently reporting an empty workspace.
- `src/commands/emoji.test.ts` (3 tests) covers the command wiring: the `list` / `get` subcommands and their `--limit` / `--no-aliases` / `--workspace` / `--json` options.
- Verified locally against a real workspace (24,980 custom emoji) under browser auth: `emoji list`, `--no-aliases`, `--limit`, `--json`, and `emoji get <name>` for both an original and an alias.
- Full suite green: `bun run type-check` clean, `bun test` 440/440 pass.

Closes #140

### DIFF
--- a/src/commands/emoji.ts
+++ b/src/commands/emoji.ts
@@ -1,0 +1,100 @@
+import { Command } from 'commander';
+import ora from 'ora';
+import { getAuthenticatedClient } from '../lib/auth.ts';
+import { error, formatEmoji, formatEmojiList, writeJson } from '../lib/formatter.ts';
+import { fetchCustomEmoji, getCustomEmoji, parseEmojiLimit } from '../lib/emoji.ts';
+
+export function createEmojiCommand(): Command {
+  const emoji = new Command('emoji')
+    .description('View a workspace\'s custom emoji');
+
+  emoji
+    .command('list')
+    .description('List the workspace\'s custom emoji')
+    .option('--limit <number>', 'Maximum number of emoji to return')
+    .option('--no-aliases', 'Exclude alias emoji, showing only originals')
+    .option('--workspace <id|name>', 'Workspace to use')
+    .option('--json', 'Output in JSON format', false)
+    .action(async (options) => {
+      const spinner = ora('Fetching custom emoji...').start();
+
+      try {
+        const client = await getAuthenticatedClient(options.workspace);
+
+        let emojiList = await fetchCustomEmoji(client, {
+          onProgress: (msg) => { spinner.text = msg; },
+        });
+
+        // `--no-aliases` sets options.aliases to false (Commander convention).
+        if (options.aliases === false) {
+          emojiList = emojiList.filter(e => !e.is_alias);
+        }
+
+        if (options.limit !== undefined) {
+          const { limit, error: limitError } = parseEmojiLimit(options.limit);
+          if (limitError !== undefined) {
+            spinner.fail('Invalid limit');
+            error(limitError);
+            process.exit(1);
+          }
+          emojiList = emojiList.slice(0, limit);
+        }
+
+        if (emojiList.length === 0) {
+          spinner.succeed('No custom emoji found');
+          return;
+        }
+
+        spinner.succeed(`Found ${emojiList.length} custom emoji`);
+
+        if (options.json) {
+          writeJson({ emoji_count: emojiList.length, emoji: emojiList });
+          return;
+        }
+
+        console.log('\n' + formatEmojiList(emojiList));
+      } catch (err: any) {
+        spinner.fail('Failed to fetch custom emoji');
+        error(err.message);
+        process.exit(1);
+      }
+    });
+
+  emoji
+    .command('get')
+    .description('Show details for a single custom emoji')
+    .argument('<name>', 'Emoji name, with or without surrounding colons')
+    .option('--workspace <id|name>', 'Workspace to use')
+    .option('--json', 'Output in JSON format', false)
+    .action(async (name, options) => {
+      const spinner = ora('Fetching custom emoji...').start();
+
+      try {
+        const client = await getAuthenticatedClient(options.workspace);
+
+        const found = await getCustomEmoji(client, name, {
+          onProgress: (msg) => { spinner.text = msg; },
+        });
+
+        if (!found) {
+          spinner.fail(`No custom emoji named :${name.replace(/^:|:$/g, '')}:`);
+          process.exit(1);
+        }
+
+        spinner.succeed(`Found :${found.name}:`);
+
+        if (options.json) {
+          writeJson(found);
+          return;
+        }
+
+        console.log('\n' + formatEmoji(found));
+      } catch (err: any) {
+        spinner.fail('Failed to fetch custom emoji');
+        error(err.message);
+        process.exit(1);
+      }
+    });
+
+  return emoji;
+}
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,7 +112,7 @@ bun run build:all          # All platforms
 
 ### Entry Point & Command Structure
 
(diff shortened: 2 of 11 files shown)


src/commands/emoji.ts
```diff
@@ -0,0 +1,94 @@
...
+          onProgress: (msg) => { spinner.text = msg; },
+        });
+
+        // `--no-aliases` sets options.aliases to false (Commander convention).
+        if (options.aliases === false) {
+          emojiList = emojiList.filter(e => !e.is_alias);
+        }
+
+        if (options.limit) {
+          emojiList = emojiList.slice(0, parseInt(options.limit));
```
vibexp-agent: `--limit abc` gives `slice(0, NaN)` -> empty list -> "No custom emoji found", and `--limit -2` silently drops the last two. Other commands hand `parseInt` straight to Slack, but here it slices locally, so a bad value looks like an empty workspace rather than an error.

Worth a validating parser on the option. Related: `src/lib/emoji.test.ts` has 10 tests, not the 12 the description claims, and neither `--limit` slicing nor `--no-aliases` filtering is among them, they only cover `src/lib/emoji.ts`. `src/commands/conversations.test.ts` is the pattern for testing this at the command level.

vibexp-agent (changes requested):
`--limit` and `--no-aliases` are the only logic this command owns and neither has a test, so a non-numeric `--limit` silently reports no emoji. Comment inline.

tmack8001:
Thanks — both are fixed in `d1eb9aa`.

**`--limit` validation (fixed).** You're right that slicing locally makes a bad value look like an empty workspace: `slice(0, parseInt("abc"))` → `[]`, and `slice(0, -2)` silently drops the tail. Rather than hand `parseInt` straight through (there's no Slack call to reject it downstream here), I added a `parseEmojiLimit` validator in `src/lib/emoji.ts` that returns `{ error: "Limit must be a positive integer" }` for anything non-numeric, negative, or zero; `emoji list` now fails with that message + exit 1 before it slices — mirroring the `canvas list` inline-validate-and-`spinner.fail` pattern already in the repo. Only applied when `--limit` is present, so the no-flag "show all" behavior is unchanged.

**Test coverage + count (fixed).** You're also right on both counts — the description said 12 but the file had 10, and neither `--limit` nor `--no-aliases` was exercised. Corrected:
- `src/lib/emoji.test.ts` now has 15 tests, adding a `parseEmojiLimit` block whose regression cases assert the exact failure modes above (`--limit abc`, `--limit -2`, `--limit 0` all rejected). I verified they're non-vacuous against the old `slice(0, parseInt(...))` path.
- New `src/commands/emoji.test.ts` (3 tests) follows the `conversations.test.ts` pattern to cover the command wiring — the `list` / `get` subcommands and the `--limit` / `--no-aliases` / `--workspace` / `--json` options.

Full suite: `bun run type-check` clean, `bun test` 440/440. The PR body is updated to match (18 tests total).

tmack8001:
One thing that needs a maintainer action before CI is meaningful here: because this is my first PR from a fork, GitHub is holding the workflow runs for approval — the `Tests`, `CI`, and `PR must link an open issue` workflows all show `action_required` ("3 workflows awaiting approval"), and the `test` check sits at *Expected — Waiting for status to be reported*. Only `check-signatures` has actually run so far.

Could you hit **Approve and run workflows** on this PR when you get a chance? Until then the real test suite hasn't executed against the branch, so a green status wouldn't yet reflect a full run. Locally it's `bun run type-check` clean + `bun test` 440/440 on `d1eb9aa`, but that's my machine, not your CI.

Happy to push again if approving surfaces anything.

vibexp-agent (commented):
Both are fixed. `parseEmojiLimit` rejects `abc`, `-2` and `0` before the slice, and the new tests are real guards: making the parser lenient again fails 3 of them, and dropping the `--limit` declaration fails the command wiring test. 440 pass and type-check clean merged with main, matching the description. Thanks for the read-only emoji group.

vibexp-agent (approved):
Verified merged with main: type-check clean, 440 pass, and the new tests are non-vacuous. Merging.

Issue #147, opened 2026-08-31
fix: throttle Slack API calls to avoid triggering unexpected_api_call_volume anomaly detection


nitrocode:
## What

Add a shared concurrency limiter / minimum inter-call delay around all Slack API requests in `SlackClient`, so slackcli's traffic doesn't burst in a way that looks like scraping to Slack's Enterprise Grid anomaly detection.

## Why

Using slackcli with browser-session (xoxc/xoxd) auth on an Enterprise Grid workspace triggered Slack's `unexpected_api_call_volume` anomalous event (see https://docs.slack.dev/reference/audit-logs-api/anomalous-events-reference/), which logged the user out. That anomaly fires when a client generates a larger volume of API calls than a browser or the official Slack client would produce.

slackcli currently has no request throttling anywhere, and several code paths fire many API calls back-to-back with no delay or concurrency cap, all funneling through `SlackClient.request()` in `src/lib/slack-client.ts`:

- `SlackClient.getUsersInfo()` calls `users.info` once per unique user ID in a tight sequential loop, no delay between calls.
- `enrichSavedItems()` (`src/lib/saved.ts`) fires a `Promise.all` that runs message-batch fetches concurrently with one `getConversationInfo` call per unique channel, all at once.
- The unread-channel resolution logic (`src/lib/unread.ts`) runs a `Promise.all` over every unread channel, each doing `getConversationInfo` (and sometimes `getUserInfo`), so concurrency scales directly with unread channel count.

None of these mirror how a real browser client paces requests, which is likely why the volume got flagged.

## How (optional)

- Add a small hand-rolled limiter (no new dependency) inside `SlackClient.request()` in `src/lib/slack-client.ts`, since every API call in the codebase already funnels through that one method (confirmed: `saved.ts`/`unread.ts` only ever call `client.*` methods, never `fetch` directly).
- Suggested defaults: max ~2 concurrent in-flight requests, minimum ~200ms between calls, applied to all requests (not just browser auth) for simplicity.
- Centralizing the gate in `request()` means `saved.ts` and `unread.ts` need no changes.
- Out of scope for this issue: retry/backoff on 429s, and the raw `fetch()` calls used for file upload/download (single-shot per command invocation, not the burst source).

vibexp-agent:
[...]

PR #152, opened 2026-09-02
feat: add --json and --message-file to messages send, edit, and draft

### LINKED ISSUE #150: feat: scriptable output for `messages send`, `edit`, and `draft` (`--json` and `--message-file`)
## WHAT — what is the feature?

Two additive flags on the message-writing commands:

**1. `--json` on `messages send`, `messages edit`, and `messages draft`**

On success, emit a single JSON object on stdout instead of the human success
line — the same `--json` shape the read commands already use:

```bash
slackcli messages send --recipient-id=C1234567890 --message="Deploying…" --json
```

```json
{
  "channel_id": "C1234567890",
  "ts": "1234567890.123456",
  "permalink": "https://myteam.slack.com/archives/C1234567890/p1234567890123456"
}
```

- `messages send` → `{channel_id, ts, permalink}`
- `messages edit` → `{channel_id, ts}`
- `messages draft` → `{channel_id, draft_id, thread_ts?}` (a draft is unsent, so
  it has no message `ts` and no permalink)

**2. `--message-file <path>` on `messages send` and `messages draft`**

Read the message body from a UTF-8 file instead of an argument, with the same
mrkdwn handling `--message` has. Mutually exclusive with `--message`.

```bash
slackcli messages send --recipient-id=C1234567890 --message-file=./release-notes.md
```

## WHY — why is it needed?

**Capturing the message identity.** Automation and AI agents rarely just post
and forget: they post a "working on it" message and then react to it, edit it
with the result, or reply in its thread. That needs `channel_id` + `ts`, and a
`permalink` to hand the message to a human or feed back into any command that
takes `--permalink`.

Today the writing commands only print a human line, so the timestamp has to be
scraped. `docs/user-guide/scripting.md` documents the workaround this forces:

```bash
# Post and keep the timestamp […] `--json` is not offered on `messages send`,
# so parse the line:
ts=$(slackcli messages send --recipient-id=C123 --message="Working…" \
     | grep -oE '[0-9]{10}\.[0-9]{6}')
```

That regex is coupled to the wording and colouring of a human-facing string, and
it silently yields an empty `ts` if the line ever changes. It also cannot
recover the permalink at all. Every read command already speaks `--json`; the
writing commands are the gap that keeps a scripted flow half-structured.

**Passing message bodies safely.** A generated message body — release notes, a
report, a rendered summary — has to survive shell quoting to reach `--message`.
Newlines, backticks, quotes, and `$` in mrkdwn make that fragile, and an
argument list has a length limit. Reading from a file removes the quoting
problem entirely, and lets an earlier pipeline step write the body directly.

It also closes a failure mode: if the generation step fails and produces an
empty file, the command should refuse rather than post an empty message.

## HOW — suggested implementation (optional)

Small and contained to `src/commands/messages.ts` plus one client method.

- **`--json`**: reuse the existing `writeJson()` helper in `src/lib/formatter.ts`
  and the same `.option('--json', …, false)` shape the read commands use, so the
  behaviour and the stdout/stderr split stay identical to `conversations read`
  and friends. Note the `writeJson()` contract (issue #73): never `process.exit()`
  after writing — the success paths here return normally, so that holds.
- **Permalink**: add `SlackClient.getPermalink()` calling `chat.getPermalink`
  (parameter is `message_ts`, not `ts`). It is a second call *after* the message
  is delivered, so a failure must not fail the send — omit the key rather than
  emitting `null`, and let the consumer test for it.
- **`--file` branch**: the external-upload flow returns the attached file rather
  than a message ts, so that branch can only offer `{channel_id, file_id}`.
- **`--message-file`**: `commander`'s `.conflicts('message')`, matching how
  `--blocks` already conflicts with `--file`. Resolve the text in one small
  exported helper so it is unit-testable, and fail there — before any Slack call —
  on a missing, empty, or whitespace-only file.

## Compatibility

This is purely additive; it mirrors the compatibility discipline of #89.

- Without `--json`, all three commands print **exactly** what they print today.
- Errors are unchanged: stderr, exit code `1`, nothing on stdout.
- `messages react` is untouched.
- `--help` for these commands gains only the new option lines.
- One internal change worth naming: `--message` can no longer be a commander
  `requiredOption` on `send`/`draft`, since `--message-file` may supply the same
  value. "Exactly one of the two" is enforced in the resolver instead, which
  keeps the missing-argument case an error — just with a message naming both
  flags. This is the same move `--permalink` already made for `--channel-id` /
  `--timestamp` on `edit` and `react`. `messages edit` takes no `--message-file`,
  so its `--message` stays required.

Documentation would be updated in the same change: `README.md`,
`docs/user-guide/messages.md`, and `docs/user-guide/scripting.md` — including
removing the `grep -oE` workaround the latter currently recommends.

## Fork

I have a working implementation on a personal fork (tests, docs, and the full CI
set green) and would be glad to open a PR once this is triaged and carries
`ready-for-pr`.

rafael-yure:
## Linked issue

Fixes #150

## Summary

Adds scriptable output to the three message-writing commands, so automation can
keep what it just wrote instead of scraping a human-readable line.

**`--json` on `messages send`, `messages edit`, `messages draft`** — replaces the
success line with one object on stdout, reusing the existing `writeJson()` helper
and the same `--json` option shape the read commands already use:

| Command | Shape |
|---|---|
| `messages send` | `{channel_id, ts, permalink}` |
| `messages send --file` | `{channel_id, file_id}` — the upload flow returns no message ts |
| `messages edit` | `{channel_id, ts}` |
| `messages draft` | `{channel_id, draft_id, thread_ts?}` — a draft is unsent, so no ts/permalink |

**`--message-file <path>` on `messages send` and `messages draft`** — reads the
body from a UTF-8 file, for long or generated text that is awkward to pass as an
argument. Conflicts with `--message` via commander; a missing, empty, or
whitespace-only file is rejected *before* any Slack call, so a failed generation
step cannot post an empty message.

This removes the workaround `docs/user-guide/scripting.md` currently recommends:

```bash
ts=$(slackcli messages send … | grep -oE '[0-9]{10}\.[0-9]{6}')
```

## What changed

- **`src/lib/slack-client.ts`** — new `getPermalink()` (`chat.getPermalink`; the
  parameter is `message_ts`, not `ts`). Narrowed `uploadFileExternal()`'s return
  from `unknown` to a small `ExternalUploadCompleteResponse` so the `--file`
  branch can read `file_id` without a cast.
- **`src/commands/messages.ts`** — two exported helpers, `resolveMessageText()`
  and `permalinkField()`, plus the new options wired into `send`, `edit`, and
  `draft`.
- **`src/commands/messages.test.ts`, `src/lib/slack-client.test.ts`** — coverage
  below.
- **`README.md`, `docs/user-guide/messages.md`, `docs/user-guide/scripting.md`** —
  documented, including replacing the `grep -oE` pattern above.

Two deliberate design points:

- **The permalink cannot fail the send.** It is a second API call *after* the
  message is delivered, so a failed lookup (a token without the scope, a
  transient error) omits the key rather than emitting `null` or failing a send
  that already succeeded. The docs tell consumers to test for the key.
- **`writeJson()` and issue #73.** That helper's contract is to never
  `process.exit()` after writing, or output past 64 KiB is dropped. These success
  paths return normally, so the contract holds.

## Compatibility

Purely additive.

- **Without `--json`, output is unchanged** on all three commands.
- Errors are unchanged: stderr, exit code `1`, nothing on stdout. Verified.
- `messages react` is untouched.
- `--help` diffs for these commands are **added lines only** — nothing removed or
  reworded.

One internal change worth calling out: `--message` is no longer a commander
`requiredOption` on `send`/`draft`, because `--message-file` can now supply the
same value. "Exactly one of the two" is enforced in `resolveMessageText()`
instead, so a missing argument is still an error — with a message naming both
flags. This mirrors what `--permalink` already did to `--channel-id` /
`--timestamp` on `edit` and `react`, and the existing test asserting the old
`mandatory` set was updated with that reasoning. `messages edit` takes no
`--message-file`, so its `--message` stays required.

## Testing

`bun test`: **433 pass, 0 fail** (was 422 — 11 added). New coverage:

- `--json` present on `send`/`edit`/`draft` and absent on `react`
- `--message-file` present on `send`/`draft` and absent on `edit`/`react`
- `--message-file` + `--message` rejected by commander on both commands
- the mandatory-option set after demoting `--message`
- `resolveMessageText`: `--message` passthrough (including a deliberately empty
  string), UTF-8 file read preserving mrkdwn/newlines/non-ASCII, and rejection of
  an empty path, a missing file, an empty file, a whitespace-only file, and
  neither-flag-given
- `permalinkField`: success, lookup throwing, lookup returning no permalink, and
  the channel/ts passed through
- `SlackClient.getPermalink`: sends `message_ts` and returns the link

Also ran the full CI set locally: `bun run type-check` clean, `bun run build`
clean, built-binary `--version` / `--help` / `auth --help` smoke tests pass,
binary 59 MB (limit 150). Pre-commit hooks installed and passing — no
`--no-verify`, no `SKIP=`. Commit is signed.

Manually verified the error paths fail before any Slack call, with exit code `1`
and an empty stdout: neither flag given, `--message` + `--message-file`, empty
file, and missing file.

## Checklist


src/commands/messages.ts
```diff
@@ -74,10 +133,12 @@ export function createMessagesCommand(): Command {
         .conflicts('file')
     )
     .option('--workspace <id|name>', 'Workspace to use')
+    .option('--json', 'Output the delivered message as JSON', false)
```
vibexp-agent, line 136: `warning()` uses `console.log`, so it writes to stdout. `warnOnWorkspaceMismatch()` runs before the send, so `messages send --permalink=<other-workspace-url> --message=x --json | jq` gets the `⚠️` line first and fails to parse. Same on `edit` and `draft`. Either skip the warning under `--json` or route `warning()` to stderr.

src/commands/messages.ts
```diff
@@ -163,6 +241,7 @@ export function createMessagesCommand(): Command {
     .option('--permalink <url>', 'Slack message link (replaces --channel-id and --timestamp)')
     .requiredOption('--message <text>', 'New message text content')
     .option('--workspace <id|name>', 'Workspace to use')
+    .option('--json', 'Output the updated message as JSON', false)
```
vibexp-agent, line 248: `--message-file` did not make it onto `edit`, so the three writing commands still take a body two different ways. `resolveMessageText()` drops in here unchanged once `--message` is demoted from `requiredOption`, same as on `send`.

vibexp-agent (changes requested):
Two things inline: the workspace-mismatch warning still lands on stdout ahead of the JSON, and `--message-file` did not make it onto `edit`. Everything else checks out.

rafael-yure:
Both addressed in 56a99dd: `warning()` now writes to stderr (pinned by a formatter test — stdout stays reserved for the `--json` object), and `messages edit` gained `--message-file` with `--message` demoted to a plain option, "exactly one of the two" enforced in `resolveMessageText` before any Slack call, same as send/draft. Docs and surface tests updated; 434 tests green.

src/commands/messages.ts
rafael-yure: Fixed in 56a99dd — warning() now writes to stderr, so stdout carries only the --json object. Pinned by a new formatter test.

src/commands/messages.ts
rafael-yure: Fixed in 56a99dd — messages edit now takes --message-file; --message demoted from requiredOption and resolveMessageText enforces exactly-one-of-two before any Slack call, same as send/draft. Docs + surface tests updated.

docs/user-guide/scripting.md
```diff
@@ -9,6 +9,10 @@ Every read command supports `--json`: `conversations read`, `conversations get`,
 `conversations unread`, `search messages`, `search channels`, `search people`,
 `saved list`, `canvas list`, `canvas read`.
 
+The writing commands support it too — `messages send`, `messages edit`, and
+`messages draft` — where it returns the identity of what was just written
+instead of the human success line.
+
 JSON goes to **stdout**. Progress spinners, error messages, and the
 update-available notice go to **stderr**, so a pipe normally carries only data:
```
vibexp-agent, line 17: Just below this, the "One caveat: the workspace-mismatch **warning** is currently written to stdout" paragraph (lines 23-26) is now false, since `warning()` goes to stderr as of this PR. Drop it.

vibexp-agent (changes requested):
Both earlier findings are properly fixed, and the `warning()` move to stderr is pinned by a test. One leftover in the docs.

docs/user-guide/scripting.md
rafael-yure: Dropped in 0d0759a. The surrounding line now lists warnings alongside spinners and errors as stderr output, so the section reads true end to end.

rafael-yure:
Rebased onto current main (through dbefb02, the shared rate limiter) — the branch was conflicting after #147/#148/#149 landed. One real conflict, in `src/lib/slack-client.ts`: main added `SlackClientOptions` and this branch added `ExternalUploadCompleteResponse` in the same spot. Both kept, nothing dropped. Docs caveat removed as noted. 478 tests green, `tsc --noEmit` clean on the rebased head 0d0759a.

vibexp-agent (approved):
Caveat is gone and the rebase kept both sides of the `slack-client.ts` conflict. Verified on top of current main: type-check clean, 479 pass, `warning()` confirmed off stdout. Merging. Nice work for a first contribution.

PR #153, opened 2026-09-02
feat: add --json output to conversations list

### LINKED ISSUE #141: feat: add JSON output to conversations list
## What

Add a `--json` option to `slackcli conversations list` so the command can return machine-readable conversation data, resolved DM users, the result count, and the pagination cursor.

## Why

`conversations list` currently rejects `--json` in Commander before any Slack API request:

```text
error: unknown option '--json'
```

This conflicts with the documented behavior:

- The README says commands that return data support `--json`.
- `docs/development/setup.md` uses `bun run dev conversations list --json` as an everyday command.
- The scripting guide lists the JSON-capable commands but omits `conversations list`.

Machine-readable output is especially important here because scripts otherwise need to parse colored, human-oriented channel and DM output. Pagination also needs structured metadata so scripts can safely request the next page.

## How

After conversations and DM users have been resolved, emit a shape like:

```json
{
  "conversation_count": 1,
  "conversations": [],
  "users": [],
  "next_cursor": "cursor-or-null"
}
```

The JSON path should return immediately after writing so the human-readable channel list and pagination command cannot contaminate stdout. The existing human-readable behavior should remain unchanged when `--json` is absent.

## Acceptance criteria

- `conversations list` exposes `--json`.
- JSON includes `conversation_count`, `conversations`, resolved `users`, and `next_cursor`.
- Empty results still produce valid JSON with an empty array and a null cursor.
- Pagination metadata is preserved in JSON.
- JSON stdout does not include the human pagination command.
- Existing human-readable output remains unchanged.
- Command tests cover both output modes and documentation describes the option.

ajayraho:
## Linked issue

Fixes #141

## Summary

Adds a `--json` flag to `conversations list`, emitting:

{
  "conversation_count": 0,
  "conversations": [],
  "users": [],
  "next_cursor": null
}

Field shape matches what @bqrkhn and @vibexp-agent settled on in the issue
thread: `SlackChannel` fields for conversations, the existing `conversations
read` shape for resolved users. Slack's empty-string `next_cursor` is
normalized to `null` in the JSON path. `writeJson()` returns immediately
after writing (no `process.exit()` call after it), so the human-readable
output and pagination hint never leak into JSON stdout — per the note about
#73 in the issue thread.

**One unrelated commit is included**, `fix: resolve findBrowser PATH scan
with platform-specific path module`. While setting up this contribution I
hit a pre-existing, Windows-only test failure in `browser-launcher.test.ts`
that blocks the mandatory local pre-commit hook (`bun test`) on Windows —
it's invisible on Linux CI since the bug only manifests when the host OS
and the simulated `platform` argument diverge. I couldn't get any commit
past the hooks without either fixing it or bypassing the hook (which the
constitution explicitly disallows), so I fixed it. Happy to pull it into
its own issue + PR instead if you'd rather keep this one strictly scoped —
just say the word.

## Checklist

- [ ] Change is focused on the linked issue — no unrelated edits *(see note above: one unrelated pre-existing-bug fix included, disclosed and open to splitting out)*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Slack API calls are throttled**: every request now goes through a process-wide rate limiter — at most 2 in flight, with at least 200ms between calls — for both standard and browser-session authentication (#147)
+  - Unthrottled bursts (one `users.info` per user, one `conversations.info` per channel) could trip Slack's `unexpected_api_call_volume` anomaly detection, which on Enterprise Grid signs the browser session out
+  - Commands that resolve many names (`saved list`, `conversations unread`) are correspondingly slower on large workspaces; the spinner keeps running while they are paced
+
 ## [0.10.0] - 2026-08-29
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -112,7 +112,7 @@ bun run build:all          # All platforms
 
 ### Entry Point & Command Structure
 
-`src/index.ts` registers the Commander.js command groups `auth`, `canvas`, `conversations`, `messages`, `saved`, `search`, and `update`. Each group is implemented in `src/commands/` and delegates to `src/lib/` modules.
+`src/index.ts` registers the Commander.js command groups `auth`, `canvas`, `conversations`, `emoji`, `messages`, `saved`, `search`, `team`, `usergroups`, and `update`. Each group is implemented in `src/commands/` and delegates to `src/lib/` modules.
 
 ### Dual Authentication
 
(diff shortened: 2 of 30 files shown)


vibexp-agent (changes requested):
This branch has none of the `--json` work: the single commit only touches `src/lib/browser-launcher.ts` and its test, and `conversations list` still has no `--json` option at 8e09274. Looks like the feature commit never got pushed.

On your question: yes please, split the browser-launcher fix into its own issue and PR. It is a real fix and worth having, just not riding along on this one.

ajayraho:
Feature commit is up now (rebuilt cleanly; had a mix-up where the commit
message and content got mismatched, fixed). Split the browser-launcher fix
into its own issue (#154) and PR (#155) as requested. This branch still
carries that fix commit for now since local hooks need it to pass on
Windows and it's not yet merged into main, once #155 merges I'll rebase
this branch to drop it, leaving just the `--json` commit.

vibexp-agent (commented):
The feature is here now and it is right: the field set matches what was settled in #141, `next_cursor` normalises to `null`, and the JSON branch returns before anything human-readable reaches stdout. Merged on main it type-checks and the suite goes 466 to 467. Thanks for this, and for splitting the browser-launcher fix out into #155.

Keep the rider commit on this branch, no need to rebase it out. It is byte-identical to #155, so whichever of the two lands first the other merges cleanly.

vibexp-agent (approved):
Verified on the updated branch: type-check clean, 467 pass, and the new command test fails if the `--json` option is removed. Merging.

PR #155, opened 2026-09-02
fix: resolve findBrowser PATH scan with platform-specific path module

### LINKED ISSUE #154: [Bug]: findBrowser PATH scan produces wrong paths on Windows
## What happened?

findBrowser() (src/lib/browser-launcher.ts) resolves a browser executable
from PATH by joining each PATH directory with a candidate name using
Node's top-level join/delimiter from 'path'. Those always follow the real
host OS rather than the platform argument findBrowser is given. This only
affects test simulation, not real usage — findBrowser's one production
call site never passes platform, so it always resolves for the real host
OS anyway, and the host-based join/delimiter were already correct there.
The practical effect is that bun test src/lib/browser-launcher.test.ts
fails on an actual Windows machine (testing findBrowser(..., 'linux') /
findBrowser(..., 'freebsd')), which blocks the mandatory pre-commit hook.
It's invisible on Linux CI because the host OS there happens to match
what most of the tests simulate.

## Steps to reproduce:

On a Windows machine, run:
bun test src/lib/browser-launcher.test.ts
"resolves a bare executable name from PATH" fails — expected
"/opt/bin/chromium", received null.

### Expected behavior

`findBrowser` should locate a PATH-only browser install on Windows the same way it does on other platforms.

### SlackCLI version

v0.10.0

### Operating system

Windows x64

### Authentication type

Both / not sure

### Relevant output or logs

```shell
src\lib\browser-launcher.test.ts:
84 |   });
85 |
86 |   it('resolves a bare executable name from PATH', async () => {
87 |     process.env.PATH = '/opt/bin';
88 |     const found = await findBrowser(existsAmong(['/opt/bin/chromium']), 'freebsd');
89 |     expect(found).toBe('/opt/bin/chromium');
                       ^
error: expect(received).toBe(expected)

Expected: "/opt/bin/chromium"
Received: null

      at <anonymous> (E:\falconry\slackcli\src\lib\browser-launcher.test.ts:89:19)
```

### Before submitting

- [x] I searched existing issues and this bug has not been reported yet.
- [x] I removed all tokens, cookies, and credentials from this report.

ajayraho:
## Linked issue

Fixes #154

## Summary

`findBrowser()` (src/lib/browser-launcher.ts) takes a `platform` parameter
so its PATH-scan behavior can be exercised for every OS from one test run,
but the PATH-scan branch used the top-level `join`/`delimiter` from
`'path'`, which always follow the real host OS rather than the `platform`
argument. `findBrowser` has exactly one production call site, and it never
passes `platform` — so in real usage `platform` is always
`process.platform`, and the host-based `join`/`delimiter` were already
correct. This does not affect `auth login-auto` on a real machine.

What it actually breaks is *simulating* another OS from a different host:
`findBrowser(..., 'linux')` or `findBrowser(..., 'freebsd')` run on an
actual Windows machine produce wrong paths, which is exactly what
`bun test src/lib/browser-launcher.test.ts` exercises — so this blocks the
mandatory pre-commit hook on Windows, without being a user-facing bug.

Fixed by selecting the `win32`/`posix` path module based on the `platform`
argument instead of the host OS, in both the implementation and its own
test, which had the identical bug in how it built its expected value.

### DIFF
--- a/src/lib/browser-launcher.ts
+++ b/src/lib/browser-launcher.ts
@@ -14,7 +14,7 @@
 
 import { spawn, spawnSync, type ChildProcess } from 'child_process';
 import { chmod, lstat, mkdir, readdir, readFile, rm, stat, writeFile } from 'fs/promises';
-import { join, delimiter } from 'path';
+import { join, posix, win32 } from 'path';
 import { homedir } from 'os';
 
 export type BrowserLaunchFailure =
@@ -135,11 +135,19 @@ export async function findBrowser(
   // PATH scan covers installs outside the standard prefixes (Nix, Flatpak
   // shims, distro variants, Scoop/Chocolatey on Windows) without hardcoding
   // every layout.
-  const pathDirs = (process.env.PATH ?? '').split(delimiter).filter(Boolean);
+  //
+  // Use the path module for the *simulated* platform, not the host's. The
+  // top-level `join`/`delimiter` always follow the real host OS, so on a
+  // real Windows machine they silently produce backslash paths even when
+  // `platform` says 'linux' or 'freebsd' (as the tests do). CI runs on
+  // Linux, where the host-OS behavior happens to match posix, so this bug
+  // was invisible there and only surfaces on an actual Windows machine.
+  const platformPath = platform === 'win32' ? win32 : posix;
+  const pathDirs = (process.env.PATH ?? '').split(platformPath.delimiter).filter(Boolean);
   const names = PATH_CANDIDATES[platform] ?? PATH_CANDIDATES.default;
   for (const name of names) {
     for (const dir of pathDirs) {
-      const full = join(dir, name);
+      const full = platformPath.join(dir, name);
       if (await fileExists(full)) return full;
     }
   }
--- a/src/lib/browser-launcher.test.ts
+++ b/src/lib/browser-launcher.test.ts
@@ -1,6 +1,6 @@
 import { afterEach, beforeEach, describe, expect, it } from 'bun:test';
 import { tmpdir } from 'node:os';
-import { join } from 'node:path';
+import { join, win32 } from 'node:path';
 import { existsSync } from 'node:fs';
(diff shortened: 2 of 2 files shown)


src/lib/browser-launcher.ts
```diff
@@ -135,11 +135,19 @@ export async function findBrowser(
...
   // every layout.
-  const pathDirs = (process.env.PATH ?? '').split(delimiter).filter(Boolean);
+  //
+  // Use the path module for the *simulated* platform, not the host's. The
+  // top-level `join`/`delimiter` always follow the real host OS, so on a
+  // real Windows machine they silently produce backslash paths even when
+  // `platform` says 'linux' or 'freebsd' (as the tests do). CI runs on
+  // Linux, where the host-OS behavior happens to match posix, so this bug
+  // was invisible there and only surfaces on an actual Windows machine.
+  const platformPath = platform === 'win32' ? win32 : posix;
```
vibexp-agent, line 145: The comment, the PR body and #154 all say this breaks `auth login-auto` on a real Windows machine. It does not. `findBrowser` has one production call site (line 262) and it passes no `platform`, so `platform` is always `process.platform`; on Windows that means the host `join`/`delimiter` were already the win32 ones and the resolved path was correct.

What this actually fixes is the simulation: `findBrowser(..., 'linux')` on a Windows host, which is why `bun test src/lib/browser-launcher.test.ts` fails there and blocks the pre-commit hook. Worth having on both counts, but the squash message will carry whatever the body says, so please reword it and #154 to the test-harness bug rather than a user-facing `login-auto` failure.

vibexp-agent (changes requested):
The change is right and the test is now a genuine guard: reverting the implementation alone makes it fail on Linux, which the old host-composed assertion could not do. The framing of the bug needs correcting though.

vibexp-agent (approved):
Taking the framing point myself rather than sending you another round trip: I am merging with a squash message that describes it as the test/simulation bug it is. Code and the new guard are right as they stand. Thanks for splitting this out.

PR #156, opened 2026-09-02
feat: add usergroups write commands with a --yes confirmation gate

### LINKED ISSUE #139: Add team and usergroups commands for workspace and user-group management

tmack8001:
## Linked issue

Closes #139

## Summary

Adds the **write** surface of the `usergroups` group, on top of the read-only commands merged in #148. This is the second and final PR of the two-PR split you ruled on for #139.

New subcommands: `create`, `update`, `add`, `remove`, `enable`, `disable`.

**`--yes` confirmation gate** (built on the existing `isInteractiveTerminal()` in `src/lib/interactive-input.ts`, per your ruling): every write subcommand confirms before acting.

- `--yes` → proceeds without prompting.
- Interactive TTY without `--yes` → prompts `y/N`.
- **stdin not a TTY and no `--yes` → REFUSES** with a clear message and a non-zero exit, rather than auto-passing — so a script, CI job, or agent cannot mutate or disable a group unattended by accident. `--yes` is the deliberate opt-in to the non-interactive path.

**`add` / `remove` are read-modify-write.** Slack's `usergroups.users.update` replaces the *entire* member list (no incremental endpoint), so the lib reads the current membership, applies the change, and writes the result back — concurrent members are never dropped, a no-op change skips the write, and it refuses to empty the last member (Slack rejects a zero-member group; disable it instead).

**Enterprise Grid:** all five write `SlackClient` methods thread an optional `team_id`; on an org, writes require `--team <T-id>` (`target_team_must_be_specified_in_org_context` otherwise).

**Auth scope:** writes need the `usergroups:write` user scope — a bot token (`xoxb`) can read but not write; use a user token (`xoxp`) or a browser session (`xoxc`/`xoxd`). Documented in `docs/user-guide/usergroups.md` alongside the `--yes` behavior.

**Tests:** lib-layer cases for `applyMembershipChange` and the add/remove read-modify-write paths (union write, no-op skip, reduced-list write, refuse-to-empty), plus a new `src/commands/usergroups.test.ts` covering `parseUserIds` splitting, `confirmWrite`'s `--yes` and non-TTY-refusal branches, and the command structure (all eight subcommands, `--yes` on the six writes). Full suite: 494 pass on this branch. Manually verified live: the non-TTY-without-`--yes` refusal (exit 1, clear message), and `usergroups --help` shows all eight verbs with `--yes` on the six writes.

## Checklist

### DIFF
--- a/docs/user-guide/usergroups.md
+++ b/docs/user-guide/usergroups.md
@@ -1,28 +1,33 @@
 # User groups
 
-`slackcli usergroups` lists and reads **user groups** — Slack's own name for a
-named, mentionable group of people (also called a "subteam", e.g.
-`@platform-team`).
+`slackcli usergroups` lists, reads, and manages **user groups** — Slack's own
+name for a named, mentionable group of people (also called a "subteam", e.g.
+`@platform-team`). This is the workspace's team-management surface.
 
 ```bash
 slackcli usergroups list
 slackcli usergroups list --include-disabled --json
 slackcli usergroups read @platform-team
+slackcli usergroups create "Platform Team" --handle=platform --description="Owns the platform" --yes
+slackcli usergroups update @platform --description="Owns platform + infra" --yes
+slackcli usergroups add @platform U0123ABC U0456DEF --yes
+slackcli usergroups remove @platform U0123ABC --yes
+slackcli usergroups disable @platform --yes
+slackcli usergroups enable @platform --yes
 ```
-
-> This PR adds the **read-only** surface (`list`, `read`). The write verbs
-> (`create`, `update`, `add`, `remove`, `enable`, `disable`) land in a
-> follow-up PR so the read side can be reviewed and merged on its own.
 
 ## Referring to a group
 
-`read` takes a `<group>` argument that accepts any of:
+Every subcommand except `list` and `create` takes a `<group>` argument. It
+accepts any of:
 
 - the group **ID** (`S03E2T070G7`),
 - the **@handle** (`@platform` or `platform`), or
 - the exact **name** (`"Platform Team"`, case-insensitive).
 
-## `usergroups list`
+## Read commands
+
+### `usergroups list`
 
 Lists the workspace's user groups, sorted by name, with each group's handle,
 member count, and enabled/disabled state.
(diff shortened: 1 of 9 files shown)


src/commands/usergroups.ts
```diff
@@ -34,9 +39,45 @@ async function requireGroup(
...
+  if (!isInteractiveTerminal()) {
+    error(
+      `Refusing to run this write unattended. stdin is not a terminal, so there is ` +
+        `no way to confirm interactively. Re-run with --yes to proceed non-interactively.`,
+    );
+    return false;
+  }
+
+  return new Promise((resolve) => {
+    const rl = readline.createInterface({ input: process.stdin, output: process.stdout });
```
vibexp-agent: `readline` writes the query to `output`, so this puts `Disable user group "X"? [y/N] ` on **stdout**, ahead of `writeJson()`. `isInteractiveTerminal()` only checks `process.stdin.isTTY`, so `slackcli usergroups disable @platform --json | jq` from a normal terminal still prompts and still breaks the pipe. `output: process.stderr` fixes it.

Separately, the comment above says a TTY "no" is a clean user cancel, but every call site does `process.exit(1)` on false, so the two cases are indistinguishable. Worth making the code and the comment agree.

docs/user-guide/usergroups.md
```diff
@@ -34,26 +39,75 @@ member count, and enabled/disabled state.
...
 
 ## Auth types
 
-Both commands go through Slack's read `usergroups.list` / `usergroups.users.list`
-methods, which work for standard (`xoxb`/`xoxp`) and browser (`xoxd`/`xoxc`)
-tokens alike.
+All of these go through Slack's `usergroups.*` methods. **Reads** (`list`,
+`read`) work for both standard (`xoxb`/`xoxp`) and browser (`xoxd`/`xoxc`)
+tokens. **Writes** need the `usergroups:write` scope, which a bot token
+(`xoxb`) does not carry — use a user token (`xoxp`) or browser session
```
vibexp-agent: `usergroups:write` is a bot scope. Slack lists its supported token types as Bot, User and Legacy Bot (https://docs.slack.dev/reference/scopes/usergroups.write), and `usergroups.create` names it for both bot and user tokens. As written this tells someone with a working `xoxb` to go get a different token.

src/commands/usergroups.ts
```diff
@@ -34,9 +39,45 @@ async function requireGroup(
   return group;
 }
 
+// Split a comma/space-separated list of user IDs into a clean array.
+function parseUserIds(raw: string[]): string[] {
```
vibexp-agent: Six new subcommands and two helpers here, and no `src/commands/usergroups.test.ts`. `confirmWrite`'s three branches and `parseUserIds`' splitting are exactly the command-layer logic that `emoji.test.ts` covers for #149. The lib tests are good but they stop at the lib boundary.

Also, the description says 11 unit cases; the diff adds 7. Suite goes 466 to 473 merged on main.

vibexp-agent (changes requested):
The read-modify-write membership handling is the right call and reads well. Three things: the prompt breaks `--json`, the bot-token claim in the docs is wrong, and the new command layer is untested.

src/commands/usergroups.ts
tmack8001: Fixed in 57ed0e9. The prompt now writes to `process.stderr` (`readline.createInterface({ input: process.stdin, output: process.stderr })`), so `usergroups <write> --json | jq` in an interactive terminal no longer has the `[y/N]` prompt land on stdout ahead of `writeJson()`. Verified live: a non-TTY `--json` invocation now emits nothing on stdout (the refusal goes to stderr).

Also reconciled the comment with the code — you're right that every call site does `process.exit(1)` on `false`, so a TTY "no" and the non-TTY refusal both decline-and-stop; the comment no longer claims a TTY "no" is a distinct "clean cancel".

docs/user-guide/usergroups.md
tmack8001: Fixed in 57ed0e9 — you're right and I verified it against the source you cited. https://docs.slack.dev/reference/scopes/usergroups.write lists the supported token types as **Bot, User, and Legacy Bot**, and names `usergroups.create/update/enable/disable/users.update` as the compatible methods. The docs now say `usergroups:write` is supported by bot, user, and legacy-bot tokens, so a bot (`xoxb`) carrying the scope can create and manage groups — the old wording wrongly sent a working `xoxb` user off to get a different token.

src/commands/usergroups.ts
tmack8001: Fixed in 57ed0e9. Added `src/commands/usergroups.test.ts` covering the command-layer logic the lib tests stopped short of:

- `parseUserIds`: comma/space splitting, leading-`@` strip + empty-drop, and the no-ids case.
- `confirmWrite`: the `--yes` branch proceeds without prompting even on a non-TTY, and the non-TTY-without-`--yes` branch refuses (returns false). `parseUserIds` and `confirmWrite` are now exported for this.
- command structure: all eight subcommands present, `--yes` on the six write verbs and absent on `list`/`read`.

And corrected the PR body's test count — it wrongly said "11 unit cases"; the suite now runs 494 pass on this branch.

vibexp-agent (approved):
All three are fixed and I verified each: the prompt now lands on stderr (stdout is empty on a piped `--json` run), the scope wording matches Slack's reference, and `src/commands/usergroups.test.ts` covers the branches. 494 pass merged on main.

I pushed one commit of my own listing the new commands and the `--yes` requirement in `docs/user-guide/scripting.md`, since an unattended caller needs to know about the refusal. Thanks for the read-modify-write membership handling, that part is genuinely well done.

Issue #175, opened 2026-09-16
Add a `channels members` command group (add / remove / list, join / leave) for channel membership management


tmack8001:
### What

A `channels members` command group to manage channel membership autonomously:

- `slackcli channels members list <channel>` — list a channel's members (`conversations.members`)
- `slackcli channels members add <channel> --users U1,U2` — add people/agents to a channel (`conversations.invite`)
- `slackcli channels members remove <channel> --users U1,U2` — remove a member (`conversations.kick`)
- `slackcli channels join <channel>` / `slackcli channels leave <channel>` — join/leave yourself (`conversations.join` / `conversations.leave`)

All accept the standard `--workspace` and `--json`. Naming mirrors both the Slack product UI ("Members" tab, "Add people or agents", "Remove from channel") and the existing `usergroups add/remove/list` convention already in the tool, so `add`/`remove`/`list` is one consistent mental model. The raw API method each wraps is noted in `--help` for the curious, but is deliberately not the CLI verb.

### Why

`slackcli` can read conversations, search, and manage user groups, but there is **no way to manage channel membership** — a common automation and offboarding need (add a new hire to their team channels, remove a departed member, script a channel's roster). Today this requires the Slack web UI. Notably, this is also a gap in the official Slack MCP server, so the browser-session-token path here reaches something app-token/MCP tooling does not.

### How (optional)

Follows the existing `docs/development/adding-a-command.md` pattern: API calls through `SlackClient.request()` (both auth types), a thin `src/commands/channels.ts`, and unit tests. `SlackClient` has **no** membership methods today (verified — no `conversations.invite/kick/join/leave/members` wrapper exists), so small new wrappers are added.

**I verified token-type acceptance live against an enterprise-grid workspace** (using a bogus channel id so nothing could mutate — a `channel_not_found` proves the token was accepted and only the channel was invalid):

| Operation | Method | Browser-token result |
|---|---|---|
| add member(s) | `conversations.invite` | `channel_not_found` → **token accepted, reachable** |
| remove member | `conversations.kick` | `channel_not_found` → **token accepted, reachable** |
| join self | `conversations.join` | `channel_not_found` → **token accepted, reachable** |
| leave self | `conversations.leave` | `channel_not_found` → **token accepted, reachable** |
| list members | `conversations.members` | `enterprise_is_restricted` → **blocked by enterprise policy** |

[...]

PR #176, opened 2026-09-17
feat(conversations): add members list for channel membership (read)

tmack8001:
## Summary

Adds a read-only `conversations members list <channel>` subcommand for inspecting channel/conversation membership, plus a `getConversationMembers` wrapper on the Slack client (calls `conversations.members`, cursor-paginated).

This is the **read half of a two-PR split** for #175. The write/self operations — `conversations members add/remove` and `conversations join/leave` — are deliberately **not** in this PR; they land in a separate follow-up PR opened only after this one merges, mirroring the read-then-write sequencing used for the usergroups arc (#139/#140).

## Maintainer decisions (vibexp-agent), and how each is met

1. **Nest under the existing `conversations` group, not a new `channels` group.** The command is `conversations members list <channel>`, added inside `createConversationsCommand()`. No new top-level group is introduced. The CLI mirrors Slack's API namespaces, and membership is `conversations.*`.

2. **Split like the usergroups arc: read first, write later.** This PR ships only `members list`. A regression test asserts the `members` group has exactly one subcommand (`list`) and that none of `add`/`remove`/`invite`/`kick`/`join`/`leave` exist, so the write verbs cannot silently leak in before their own PR.

3. **Degrade honestly on `enterprise_is_restricted`.** On an Enterprise Grid, `conversations.members` returns `enterprise_is_restricted` regardless of team scoping (verified live: `--team` does not lift it). The command still **exists** and, on that error, prints a clear human-readable message explaining member enumeration is blocked by org policy and that team scoping does not lift it, then exits non-zero. It does not hide the command or fake success. In `--json` mode it fails through the same clear-error, non-zero-exit path.

## What changed

`src/lib/slack-client.ts` gains `getConversationMembers(channel, { cursor, limit })`, a thin read wrapper over `conversations.members`. It adds no write methods.

`src/commands/conversations.ts` gains a `members` sub-group with a single `list` subcommand. It accepts a channel ID **or** a Slack link (reusing the existing `parseSlackLink` permalink helper), supports `--limit` with members-returned semantics (consistent with the CLI's other `--limit` flags), `--cursor` pagination, `--workspace`, and `--json`. JSON output shape is `{ channel_id, member_count, members, next_cursor? }`, mirroring the existing `conversations read` JSON shape, and uses the `writeJson` path (no `process.exit` after it, per the issue #73 guidance in that helper).

## Tests

Scoped tests only (added, all green):

- `src/lib/slack-client.test.ts`: `getConversationMembers` calls `conversations.members` with the channel, passes the pagination cursor through, and surfaces `enterprise_is_restricted` (does not swallow it).
- `src/commands/conversations.test.ts`: `members list` is nested under `conversations`, requires the `<channel>` positional, exposes `--limit`/`--cursor`/`--workspace`/`--json`, and ships ONLY the read side (guards against write verbs leaking in).

`bun run type-check` passes; the pre-commit hook (type check + tests) passed on commit.

## Read-only confirmation

No write API methods and no write subcommands are introduced. The only Slack method added is the read call `conversations.members`.

## Follow-up

The write/self membership operations are intentionally deferred to a second PR after this merges, per the maintainer's read-first split. Both PRs link back to #175.

Part of #175
Refs #175


src/commands/conversations.ts
```diff
@@ -343,5 +363,88 @@ export function createConversationsCommand(): Command {
...
+    .option('--limit <number>', 'Maximum number of members to return', '100')
+    .option('--cursor <cursor>', 'Pagination cursor for next page of results')
+    .option('--workspace <id|name>', 'Workspace to use (overrides default)')
+    .option('--json', 'Output in JSON format', false)
+    .action(async (channelArg, options) => {
+      const spinner = ora('Fetching members...').start();
+
+      try {
+        const { channelId, workspace } = resolveChannelArg(channelArg);
+        const limit = parseInt(options.limit);
```
vibexp-agent: `parseInt(options.limit)` isn't validated before it feeds the pagination loop and `memberIds.slice(0, limit)` below. `--limit abc` parses to `NaN`, the `while` loop never runs (`0 < NaN` is false), and this prints "Found 0 members" as a success instead of an error. `users list` in #177 validates with `Number.isFinite(limit) && limit > 0` before proceeding; worth doing the same here.

vibexp-agent (changes requested):
Good split, and the enterprise_is_restricted handling is exactly right. One bug before this can merge, plus docs/user-guide/conversations.md (and the scripting.md --json list) need a section for `members list` like the other subcommands have.

src/commands/conversations.ts
```diff
@@ -377,5 +397,88 @@ export function createConversationsCommand(): Command {
...
```
vibexp-agent: `parseInt(options.limit)` isn't validated before it feeds the pagination loop and `memberIds.slice(0, limit)` below. `--limit abc` parses to `NaN`, the `while` loop never runs (`0 < NaN` is false), and this prints "Found 0 members" as a success instead of an error. `users list` in #177 validates with `Number.isFinite(limit) && limit > 0` before proceeding; worth doing the same here.

vibexp-agent (changes requested):
Re-posting against the current head after pushing a conflict-only merge with main (getTeamInfo/usergroups/emoji.list additions in slack-client.ts) so CI can actually run here; no change to this PR's own code. The finding below still stands.

tmack8001:
Addressed both findings in `fa7291b` (pushed on top of `d0229e8`).

**✅ Bug — unvalidated `--limit` (`conversations.ts` members list)**

Fixed. `--limit` is now parsed with `parseInt(options.limit, 10)` and rejected before the pagination loop starts:

```ts
const limit = parseInt(options.limit, 10);
if (!Number.isFinite(limit) || limit <= 0) {
  error('--limit must be a positive integer');
  process.exit(1);
}
```

This mirrors the `Number.isFinite(limit) && limit > 0` guard `users list` uses in #177. `--limit abc` now exits non-zero with a clear message instead of printing "Found 0 members" as a false success. Reproduced the old behavior before the fix (NaN made `0 < NaN` false, so the `while` never ran) and confirmed the guard catches it.

**✅ Docs — `members list` section**

Added:
- `docs/user-guide/conversations.md` — a `conversations members list` section matching the other subcommands' style: examples (basic / `--limit` / `--cursor` / `--json`), the members-returned semantics, the `--json` shape `{ channel_id, member_count, members, next_cursor? }`, and the `enterprise_is_restricted` caveat (noting `--team` does not lift it and that add/remove/join/leave ship separately).
- `docs/user-guide/scripting.md` — added `conversations members list` to the `--json` read-command enumeration.

**Test / gates**

Added a structural test asserting the wired `--limit` default passes the same guard the action enforces (matching the file's existing wiring-test style; #177's validation shipped with structural tests too). Full suite 521/521 green, type-check clean, committed through the pre-commit hook.

vibexp-agent (approved):
Limit validation and docs both look right, and I confirmed `--limit abc` rejects with exit 1. Squash merging.

PR #177, opened 2026-09-17
feat(users): add users command group with shared profile-field resolver

### LINKED ISSUE #174: Add a `users` command group (users info / users list) with account status and --resolve-fields
### What

A new `users` command group to look up individual users and list workspace members, surfacing the fields that `search people` can't:

- `slackcli users info <id>` — fetch one user via `users.info`, including **account status** (`deleted` / deactivated), timezone (`tz`, `tz_label`, `tz_offset`), title, email, and admin/guest flags.
- `slackcli users list` — list workspace members via `users.list` (cursor-paginated), with a `--status all|active|deactivated` filter.
- `--resolve-fields` (on both) — resolve the opaque custom profile-field IDs (`XfXXXXXXXX`) to their human labels via `team.profile.get`.

All accept the standard `--workspace` and `--json`.

### Why

`slackcli` can search for a person (`search people`) and list user groups (`usergroups`, from #139), but there is **no way to answer three common questions about a specific user**:

1. **Is this account deactivated?** The `search.modules` people endpoint that backs `search people` does **not** return the `deleted` flag — so you cannot tell from its JSON whether someone has been offboarded, even though the Slack UI clearly shows a "Deactivated account" banner. The `deleted` boolean only comes back from `users.info` / `users.list`.
2. **What is their timezone / working context?** `tz`, `tz_label`, `tz_offset` are absent from `search people` and only present on `users.info`.
3. **What do the custom profile fields mean?** `profile.fields` keys are opaque workspace-defined IDs; their labels live in a separate `team.profile.get` definitions table.

These are natural offboarding/audit/automation needs (e.g. "which members in this list are deactivated?"), and today they require dropping to a raw Slack API call. The underlying methods (`users.info`, `users.list`, `team.profile.get`) all work under `slackcli`'s existing dual-auth model, so no new auth surface is introduced.

### How (optional)

Follows the existing `docs/development/adding-a-command.md` pattern: API calls routed through `SlackClient.request()` (both auth types), a thin command file in `src/commands/users.ts`, formatters, and unit tests. `SlackClient` already has `getUserInfo` (`users.info`), `getUsersInfo` (batch), and `listUsers` (`users.list`); only a small `getTeamProfile` (`team.profile.get`) wrapper is new.

I built a working prototype off `main` and verified all three behaviors live against a real workspace. Findings that shaped the design:

- **`deleted` is the reliable status signal**, verified live on both `users.info` and `users.list` (in a `users.list` page of 600 members, all 600 carried a `deleted` boolean). Example (`users info` on a deactivated user, all values fabricated):

  ```
  $ slackcli users info U0EXAMPLE01
    Name:     Jane Example
    Handle:   @jexample
    ID:       U0EXAMPLE01
    Email:    jane.example@example.com
    Title:    Staff Engineer
    Status:   DEACTIVATED
    Deleted:  true
    Admin:    false
    TZ:       (none) (n/a, offset n/a)
  ```

- **A deactivated user's `users.info` is stripped down** — verified live: for a deactivated account `tz`/`tz_label`/`tz_offset` and `is_admin` are *omitted*, whereas an **active** user returns them populated. The two observed shapes (values fabricated, structure real):

  ```
  active     → { "deleted": false, "tz": "America/Somewhere", "tz_label": "Central Time", "tz_offset": -18000, "is_admin": false }
  deactivated → { "deleted": true }   // no tz / tz_label / tz_offset / is_admin keys
  ```

  So the command must treat `tz`/`is_admin` as optional and rely on `deleted` as the always-present status field.

- **`search people` genuinely omits these fields** — verified live against the same active user that `users.info` returned a timezone for: its `search people` object has **no `deleted` key and no `tz` key** (neither top-level nor under `profile`); its top-level keys are `iid, id, profile, username, is_restricted, is_ultra_restricted, phone, team_id, enterprise_user, top_terms, similar, timestamp`. This is the core gap this command closes.

- **`--resolve-fields` works** — it maps IDs to labels via one `team.profile.get` call (cached per invocation). Labels are admin-renameable, so this is resolved live, not hardcoded. Example (fabricated):

  ```
  $ slackcli users info U0EXAMPLE02 --resolve-fields
    Fields:
      Country: US
      Department: Platform Engineering
      Organization: Infrastructure
      Employee Type: Regular
  ```

- **`users list` paginates and scales** — cursor-loops `users.list` and the status filter works across a full workspace (verified on thousands of members).

**Open question for the maintainer (scan-depth vs. matches-returned semantics):**
`users.list` has **no server-side status filter** — Slack returns all members in a fixed order, so `--status` must be applied client-side (fetch-then-filter). That makes `--limit` ambiguous. In the prototype, `--limit N` means "fetch up to N members, *then* filter", so `--status active --limit 5` can return very few matches if the first pages happen to be bots/deactivated. Which semantics do you prefer?

- **(a)** `--limit` = scan depth (fetch N, then filter) — current prototype behavior; cheap, but a small limit can return few/zero matches.
- **(b)** `--limit` = matches returned (keep paginating until N *matching* users are collected, or the workspace is exhausted) — more intuitive, but can scan the whole workspace for a rare status.
- **(c)** two separate flags, e.g. `--scan <N>` (pages to fetch) and `--limit <N>` (matches to return).

Happy to implement whichever you choose. I have a signed, tested prototype branch ready and will open the PR once this is triaged with `ready-for-pr`.

**Second question — should `--resolve-fields` be shared with `search people`?**
`search people`, `users info`, and `users list` all return the same `profile.fields` shape, so the resolver is really orthogonal to the command that produced the user. I'd propose implementing it as a shared `src/lib/` helper (`resolveProfileFields(user, labelMap)`) that all three call, and adding `--resolve-fields` to `search people` too, so its output becomes human-readable without a separate `team.profile.get` step.

Important scope boundary I'd suggest keeping: `--resolve-fields` should mean **label** resolution only (field ID → label, e.g. `XfXXXXXXXX` → `Department`) — one cached `team.profile.get` call, O(1) regardless of user count. It should **not** resolve field *values*: `Manager`/`Direct Reports` are type `user` (their value is a `U…` ID), and turning those into names needs a `users.info` per referenced user — an O(N) fan-out on `users list` that re-introduces the `unexpected_api_call_volume` throttling concern addressed in #147. If value resolution (manager/team IDs → names) is wanted, I'd suggest a separate, explicitly opt-in flag (e.g. `--resolve-users`) so the cheap label case stays cheap. Open to your call on whether to include the `search people` port in this PR or split it.

**Out of scope:** changing workspace membership (`admin.users.*`) — that needs an admin/enterprise token browser session tokens can't provide; this is read-only.

tmack8001:
## Summary

Adds a `users` command group — `slackcli users info <id>` and `slackcli users list` — for looking up a user by ID and enumerating workspace users by account status, and ships a shared `--resolve-fields` resolver across `users info`, `users list`, and `search people`.

Closes #174

## Motivation

`search people` finds users by name/email, but there was no way to look up a single user by ID or to enumerate the workspace filtered by account status (active vs deactivated). Custom profile fields also came back only as opaque `Xf…` IDs, unreadable without a label map.

## Maintainer decisions, as implemented

**(1) `--limit` = matches-returned.** `users list --limit N --status active` returns up to N users that PASS the status filter. It pages `users.list` internally, applying the filter as it goes, and stops when it has N matches or the workspace is exhausted — never a scan-depth cap. This is what `--limit` means everywhere else in the CLI (`search`, `conversations list`). Lives in `src/lib/users.ts` (`listUsersByStatus`) so it is unit-testable without the network.

**(2) `--resolve-fields` as a shared `src/lib/` helper, wired into all three surfaces in this PR.** `src/lib/profile-fields.ts` exposes `buildFieldLabelMap` (one cached `team.profile.get` call) and `resolveProfileFields(user, labelMap)`. `users info`, `users list`, and `search people` all call it, so the flag behaves identically wherever it appears. It resolves IDs to their human labels only — an O(1) label lookup, no per-user fan-out.

**(3) No `--resolve-users`.** Field values that are themselves user IDs (Manager, Direct Reports) are passed through as `U…` IDs, not names. Resolving those to names is a second, per-referenced-user hop; it is intentionally out of scope here and left to a separate future issue/PR. The boundary is stated in the help text and the user guide.

## What changed

Problem: no by-ID lookup, no status enumeration, and custom fields were unreadable IDs.

Approach: a thin command layer over `users.info` / `users.list` / `team.profile.get`, with the paging and resolving logic pushed into `src/lib/` per the repo's adding-a-command convention, and the field resolver shared rather than duplicated.

Change: new `users` group (`info`, `list`); `--resolve-fields` added to `search people` using the same resolver; `--limit` on `users list` switched to matches-returned; docs and tests added.

Account status is derived from `deleted`, the one deactivation signal Slack always returns — a deactivated user has `tz`/`is_admin` stripped, so those cannot be relied on for status. `--status` filters on that axis: `active` = not deactivated (bots included), `deactivated` = `deleted:true`, `all` = both; the per-user `status` field still distinguishes bot/guest/active.

## Tests

Scoped, offline, no live credentials — following the repo's `SlackClient`-subclass pattern.

`src/lib/users.test.ts` — matches-returned semantics: keeps paging past all-non-matching pages until N matches, stops early once N is reached (asserting only one page was fetched), returns fewer than N when the workspace is exhausted, counts deactivated matches, counts bots as active, and surfaces a not-ok `users.list` as a thrown error.

`src/lib/profile-fields.test.ts` — label map built from one `team.profile.get` call, raw-ID fallback for an unlabelled field, user-typed values left as `U…` IDs (the decision-3 boundary), unknown IDs kept not dropped, empty/missing values normalised.

`src/commands/users.test.ts` — `--resolve-fields` is registered on `users info`, `users list`, and `search people`; `users list` carries `--limit`/`--status`.

The full `bun test` suite and `bun run type-check` pass via the pre-commit hook.

## Manual verification

The prototype was live-verified against a real workspace earlier this cycle: `users info` on a deactivated account reported `DEACTIVATED`/`Deleted true`; `--resolve-fields` labelled the opaque `Xf…` IDs; `users list --status deactivated` enumerated the deactivated set. The matches-returned refactor and the shared-resolver extraction are covered by the offline unit tests above.

## Notes

I cannot apply labels on this fork PR. Suggested for triage: `enhancement`. The `ready-for-pr` gate for #174 was satisfied by the maintainer's decisions on the issue.


vibexp-agent (approved):
This is well done: matches-returned pagination, the shared field-label resolver wired into all three surfaces consistently, and the labels-only boundary is tested and honest. No code findings.

Pushed a conflict resolution against main (team/usergroups landed since this branch was cut) plus two mechanical doc gaps: users info/list into scripting.md's --json list, and the required-scopes line in users.md matching files.md's convention. Also registered users.md in web/docs.manifest.mjs, which the site build silently required.

PR #178, opened 2026-09-18
feat(conversations): channel-membership write and self ops (add/remove/join/leave)

### LINKED ISSUE #175: Add a `channels members` command group (add / remove / list, join / leave) for channel membership management

tmack8001:
## Linked issue

Closes #175

## Summary

You can now change who is in a channel from the CLI.

Four commands, filling in the group whose read-only `members list` already shipped:

- `conversations members add <channel> <users...>` — put people (or agents/apps) in a channel.
- `conversations members remove <channel> <users...>` — take people out.
- `conversations join <channel>` — join a public channel as yourself.
- `conversations leave <channel>` — leave a channel as yourself.

The names match what Slack itself says on the Members tab — "Add people or agents", "Remove from channel" — and match this CLI's own `usergroups add`/`remove`, so it's one mental model everywhere.

### How each one behaves

**add** hands the whole user list to Slack's `conversations.invite` in one call. That call is all-or-nothing: if even one id can't be invited, Slack invites nobody and returns an error, which the command surfaces — so a success message never lies about who got added.

**remove** is different because Slack's `conversations.kick` takes one user at a time. So the command loops, and it's best-effort: it tries every id and tells you which came out and which didn't, instead of stopping at the first failure. If any removal fails, the command exits non-zero (so a script notices) and the `--json` output carries the full `removed` / `failed` split.

**join** asks for no confirmation — joining is harmless and joining a channel you're already in does nothing. **leave** does ask (with `--yes` to skip), because leaving is a real change to your membership; leaving a channel you're already out of is reported as a no-op, not an error.

Every command that *changes* something (add, remove, leave) goes through the same confirmation gate the `usergroups` write commands use: in a real terminal it prompts, and when there's no terminal it refuses unless you pass `--yes` — so a script can't mutate membership by accident. Prompts go to stderr, so `--json` output on stdout stays clean.

```mermaid
flowchart LR
  subgraph before["Before — read only"]
    L1["members list"]:::ctx
  end
  subgraph after["After — this PR"]
    L2["members list"]:::ctx
    A["members add"]:::added
    R["members remove"]:::added
    J["join"]:::added
    LV["leave"]:::added
  end
  classDef added fill:#DCFCE7,stroke:#16A34A;
  classDef ctx fill:#E0F2FE,stroke:#0284C7;
```

🟩 added · 🟦 unchanged — the `members` group gains `add`/`remove`, and `join`/`leave` land on the `conversations` group as self-ops.

### Enterprise note

Only member *enumeration* (`members list`) is blocked by Enterprise Grid policy (`enterprise_is_restricted`). The write and self ops are reachable and are not special-cased — they surface Slack's own error codes through the normal path.

### Docs

`docs/user-guide/conversations.md` gains a section per new command (examples, confirmation/`--yes` behavior, `--json` shape), and updates the `members list` cross-reference now that management and self ops are documented rather than "shipping separately". `docs/user-guide/scripting.md` lists the four new verbs among the writing commands that support `--json`.

## Checklist

- [x] Documentation in `docs/` updated — `conversations.md` gains sections for `members add`/`remove` and `join`/`leave`; `scripting.md` lists them under the writing commands that support `--json`

## Tests

`src/commands/conversations.test.ts` — the read PR's guard test that asserted the write verbs were *absent* is flipped to assert they now exist, plus structural tests for the arg shapes, the `--yes` split (present on add/remove/leave, absent on the idempotent join), and the `--team` split (write ops only, not self-ops). 549/549 tests pass, type-check clean.

## Manual verification

The four Slack methods were probed live during issue #175's investigation: `conversations.invite`/`kick`/`join`/`leave` all accept the browser token (return `channel_not_found` on a bogus channel = reachable), and `conversations.members` is the only one blocked by `enterprise_is_restricted`. `conversations.invite`'s atomic-batch behavior was re-verified against Slack's published API docs.


src/lib/slack-client.ts
```diff
@@ -417,6 +417,37 @@ export class SlackClient {
...
+  // Join a public channel as the authenticated user. conversations.join is a
+  // self-op (no target user) and returns the joined channel object.
+  async joinConversation(channel: string): Promise<any> {
+    return this.request('conversations.join', { channel });
+  }
+
+  // Leave a conversation as the authenticated user. conversations.leave is a
+  // self-op; Slack returns { ok, not_in_channel? } where not_in_channel=true
+  // means we were already out (a no-op, not a failure).
+  async leaveConversation(channel: string): Promise<any> {
```
vibexp-agent, line 467: `request()` throws on `!data.ok` for both auth paths (the `@slack/web-api` `apiCall` throws on `result.ok === false`, and `browserRequest` does the same), and Slack's `not_in_channel` response comes back with `ok:false`. So this always rejects when you're already out of the channel, it never resolves with `{ not_in_channel: true }`.

src/commands/conversations.ts
```diff
@@ -485,5 +488,188 @@ export function createConversationsCommand(): Command {
...
+      if (!(await confirmWrite(`Leave ${channelId}?`, options.yes))) {
+        process.exit(1);
+      }
+      const spinner = ora('Leaving channel...').start();
+      try {
+        const client = await getAuthenticatedClient(options.workspace);
+        warnOnWorkspaceMismatch(client, workspace);
+
+        const response = await client.leaveConversation(channelId);
+        const alreadyOut = response.not_in_channel === true;
```
vibexp-agent, line 660: Dead code: `leaveConversation` throws before this line runs whenever you're already out (see slack-client.ts), so `leave` on a channel you're not in falls into the catch block and exits 1 with a raw Slack error, not the no-op success described here and in the docs. Needs to catch that specific case (e.g. inspect `err`'s underlying data for `not_in_channel`, or treat it as success) rather than relying on a response field that's unreachable.

vibexp-agent (changes requested):
Nice work splitting add/remove/join/leave out cleanly. One bug in `leave` before this can merge.

tmack8001:
Thanks for the catch — this was a real bug, and digging in showed it was worse than a one-liner.

**✅ Fixed in `25f7420`.**

You're right that `leave` never reached its no-op-success path. Verifying against [Slack's `conversations.leave` docs](https://docs.slack.dev/reference/methods/conversations.leave), the already-out response is `{ "ok": false, "not_in_channel": true }` — with **no `error` field**. Since `request()` throws on `ok:false` for both auth paths, and neither path surfaces `not_in_channel` in the thrown message (browser: `data.error` is undefined → `"Unknown API error"`; standard: `@slack/web-api` message is `"An API error occurred: undefined"`), the already-out case fell into the catch and exited 1 with a raw error — exactly as you described.

The fix is structural rather than a message-string match (which couldn't work, since the discriminator isn't in the message):

- Both transports now preserve the full Slack payload on the thrown error as `.slackData` (`@slack/web-api` already exposes it via `error.data`; `browserRequest` now attaches it too). This is additive — the error message is unchanged, so no other caller is affected.
- `leaveConversation` inspects `err.slackData.not_in_channel` and returns the payload as the no-op success Slack intends, re-throwing any genuine error (e.g. `cant_leave_general`).
- The command's `response.not_in_channel === true` success branch is now reachable and correct.

Added two regression tests in `slack-client.test.ts` exercising the real response shapes through the actual fetch path: already-out returns the payload (no throw); a genuine error still throws.

Verified: `tsc --noEmit` clean, full suite **551/551 green** through the pre-commit hook. Ran an adversarial review-to-consensus pass before pushing — the first message-match attempt was caught as a no-op and replaced with this structural fix.

vibexp-agent (approved):
Fix checks out. `leaveConversation` now recovers `not_in_channel` off `.slackData` before it hits the catch, both transports preserve it through the re-wrap, and the new tests are real (verified by reverting the fix locally: it breaks exactly the added no-op test). 551/551 green.

Issue #184, opened 2026-09-20
refactor(saved): reduce cognitive complexity of enrichSavedItems (S3776, 52/15)


vibexp-agent:
## What
SonarCloud flags `typescript:S3776` (Cognitive Complexity of functions should not be too high) in:
- `src/lib/saved.ts:4` — Refactor this function to reduce its Cognitive Complexity from 52 to the 15 allowed.

`enrichSavedItems()` is a single ~135-line function that does five unrelated jobs: cursor pagination over `listSavedItems`, branching on the browser-vs-standard response shape, grouping timestamps per channel and chunking them into batched `listMessages` calls, building the channel-name and message lookups, and finally resolving user info.

## Why
At 52 this is by a wide margin the most complex function in the codebase (the rule's limit is 15, and the next worst is 35). The browser and standard auth paths are interleaved inside one body, which is exactly the seam `CLAUDE.md` says must stay clear — `slack-client.ts` dispatches on `auth_type`, but here the two shapes are re-discovered from the response (`isBrowserFormat`) and handled inline. That makes the function hard to test per-path and easy to break when one auth mode's response shape changes.

## How
Extract along the seams that already exist in the code, keeping `enrichSavedItems()` as a thin orchestrator:
- `fetchAllSavedItems(client, options)` — the pagination `do/while`, returning `{ rawItems, isBrowserFormat }`.
- `enrichBrowserItems(client, rawItems, onProgress)` — the batching, channel-info and message-lookup block.
- `enrichStandardItems(rawItems)` — the `stars.list` branch.
- `fetchUsers(client, userIds, onProgress)` — the trailing user lookup.

Behaviour must not change; add tests per extracted helper (`src/lib/saved.test.ts`), covering both auth shapes, the `options.limit` truncation, and the `[message unavailable]` fallback.

---
Filed automatically from SonarCloud triage. Not yet authorized for
implementation — this issue needs the maintainer's `ready-for-pr` label first
(CLAUDE.md §2).

---
## WHAT
**Summary:** Split `enrichSavedItems()` into four single-purpose helpers so no function exceeds SonarCloud's cognitive-complexity limit of 15.

**Scope:** `src/lib/saved.ts` only — pagination, browser-format enrichment, standard-format enrichment and user lookup each become their own function; `enrichSavedItems()` stays the exported orchestrator.

**Affected Components:**
[...]

PR #195, opened 2026-09-20
refactor(mrkdwn): extract parseInline marker matching into helpers (#185)

### LINKED ISSUE #185: refactor(mrkdwn): reduce cognitive complexity of parseInline (S3776, 35/15)
## What
SonarCloud flags `typescript:S3776` (Cognitive Complexity of functions should not be too high) in:
- `src/lib/mrkdwn.ts:56` — Refactor this function to reduce its Cognitive Complexity from 35 to the 15 allowed.

`parseInline()` nests a `for (const [marker, styleKey] of MARKERS)` scan inside a `while (i < text.length)` walk, with six `continue` guards (marker mismatch, word boundary before, no closing marker, word boundary after, empty inner, space-padded inner), a recursive branch for nested styles, a `matched` flag driving the plain-character fallback, and a trailing `map` that strips empty style objects.

## Why
This is the function that turns user text into Slack `rich_text` blocks for `drafts.create`, so a regression here silently corrupts every draft the CLI writes. At complexity 35 the marker-matching rules — particularly the `_`-only word-boundary logic that the inline comments carefully explain — are buried inside the loop, so a future change to one marker's rules has to be reasoned about against the whole walk. The `matched` boolean coupling the inner loop to the outer one is the specific structure making this hard to follow.

## Why not a sweep
`typescript:S3776` is never bulk-fixed: reducing cognitive complexity is a bespoke redesign each time, and a botched split changes behaviour silently. This issue is scoped to one file.

## How
Extract the inner marker attempt into a pure helper and let the outer loop stay a simple walk:
- `tryMatchMarker(text, i): { element(s), nextIndex } | null` — owns the six guards and the recursion, returning `null` when no marker applies at `i`.
- `appendPlainChar(elements, char)` — the plain-character accumulation.
- `stripEmptyStyles(elements)` — the trailing cleanup `map`.

`parseInline()` then reduces to: try a marker, else append a plain char, advance. The existing test suite in `src/lib/mrkdwn.test.ts` must pass unchanged, and new cases should pin the boundary rules the comments describe (`file_name` not italicising, `*bold*` mid-word, unterminated markers left literal, nested `*_x_*`).

---
Filed automatically from SonarCloud triage. Not yet authorized for
implementation — this issue needs the maintainer's `ready-for-pr` label first
(CLAUDE.md §2).

---
## WHAT
**Summary:** Extract the marker-matching attempt out of `parseInline()` into a pure helper so the mrkdwn walker drops under the cognitive-complexity limit of 15.

**Scope:** `src/lib/mrkdwn.ts` only. `parseInline()` and `parseMrkdwn()` keep their signatures and output; the six marker guards move into one helper.

**Affected Components:**
- `src/lib/mrkdwn.ts` — add `tryMatchMarker()`, `appendPlainChar()`, `stripEmptyStyles()`; shrink `parseInline()`.
- `src/lib/mrkdwn.test.ts` — extend.

**Acceptance Criteria:**
- [ ] SonarCloud reports no `typescript:S3776` finding in `src/lib/mrkdwn.ts` (was 35, limit 15).
- [ ] The existing `src/lib/mrkdwn.test.ts` suite passes **unchanged** — no test edited to accommodate the refactor.
- [ ] New tests pin the rules the current inline comments describe: `file_name` does not italicise, `*bold*` applies mid-word, an unterminated marker is left literal, `_ spaced _` is not a span, nested `*_x_*` merges both styles, and a code span does not nest.
- [ ] `bun run type-check` and `bun test` pass.

**Out of Scope:** adding mrkdwn features (links, lists, blockquotes, user mentions); changing which markers are supported; altering the `rich_text` block shape emitted by `parseMrkdwn()`.

## WHY
**Business Value:** This function converts user text into the `rich_text` payload for `drafts.create`. It is the only writer of that format, so a regression corrupts every draft the CLI creates — and drafts are a browser-auth-only path that is awkward to spot-check manually.

**Problem Statement:** `parseInline()` nests a `for (const [marker, styleKey] of MARKERS)` scan inside a `while (i < text.length)` walk, with six `continue` guards, a recursive branch for nested styles, and a `matched` boolean coupling the inner loop back to the outer one. At complexity 35 the delicate part — the `_`-only word-boundary rule, which existing comments explain at length because getting it wrong swallows text between two unrelated underscores — is not separable from the walk.

**Expected Impact:** Cognitive complexity 35 → under 15. The marker rules become independently testable, which is what makes the boundary behaviour safe to change later.

**Context:** Filed from SonarCloud triage; `AaC6wGpICitTlL4xdYTY`. `typescript:S3776` is deliberately never bulk-fixed — each reduction is a bespoke redesign — so this issue covers one file. Peer issues from the same run: #184, #186, #187 — independent, no ordering.

## HOW
**Technical Approach:** Invert the control flow. Instead of an inner loop setting a `matched` flag that the outer loop reacts to, have a helper return either a match result or `null`, and let the outer loop be a plain "try a marker, else take one plain character, advance" walk. Behaviour-preserving: the helper keeps the same guard order, since the guards are order-sensitive.

**Implementation Details** — all in `src/lib/mrkdwn.ts`:
- `function tryMatchMarker(text: string, i: number): { elements: RichTextElement[]; nextIndex: number } | null` — iterates `MARKERS`, applies all six guards in their current order (marker mismatch → `wordCharBefore` for `_` → no closing marker → `wordCharAfter` for `_` → empty inner → space-padded inner), and owns both the code-span branch and the recursive `parseInline(inner)` style-merge. Returns `null` when nothing matches at `i`.
- `function appendPlainChar(elements: RichTextElement[], char: string): void` — the append-or-create-last-plain-element logic.
- `function stripEmptyStyles(elements: RichTextElement[]): RichTextElement[]` — the trailing `map` that removes empty `style` objects.
- `parseInline()` becomes roughly: `while (i < text.length) { const m = tryMatchMarker(text, i); if (m) { elements.push(...m.elements); i = m.nextIndex; } else { appendPlainChar(elements, text[i]); i++; } } return stripEmptyStyles(elements);`
- Leave `wordCharBefore`/`wordCharAfter` and their surrogate-pair handling untouched.

**Testing Strategy:** Unit tests in `src/lib/mrkdwn.test.ts` — pure functions, no fakes needed. Test `tryMatchMarker()` directly for each guard, and keep end-to-end `parseMrkdwn()` assertions for the emitted block shape. The existing suite is the regression net; treat any need to edit it as a sign the refactor changed behaviour.

**Documentation / Configuration:** No user-facing change; no `docs/` or `CHANGELOG.md` entry. No new dependencies.

**Files to Create / Modify:** Modify `src/lib/mrkdwn.ts`, `src/lib/mrkdwn.test.ts`.

**Risks:**
- Guard order is load-bearing — reordering changes what parses → keep the exact sequence and assert each guard individually.
- The recursion currently calls `parseInline(inner)`, which will now route through the new structure; make sure `tryMatchMarker` still recurses into `parseInline` (not into itself) so nested style merging is unchanged.
- Surrogate-pair handling in `wordCharBefore` is subtle; do not inline or "simplify" it while moving code around.

**Complexity:** Medium (2 files, <500 LOC, behaviour-critical)

---
**Refined by:** AI Issue Refiner · **Date:** 2026-09-21 · **Status:** Ready

vibexp-agent:
Closes #185

## What

Splits `parseInline()` in `src/lib/mrkdwn.ts` into single-purpose module-private helpers so SonarCloud `typescript:S3776` drops from 35 to well under the limit of 15. Pure extraction — no behaviour change.

- `findSpan(text, marker, i)` — owns the six order-sensitive guards (marker mismatch → `wordCharBefore` for `_` → no closing marker → `wordCharAfter` for `_` → empty inner → space-padded inner), returning `{ inner, nextIndex }` or `null`.
- `styleSpan(inner, styleKey)` — the code-span branch and the recursive `parseInline(inner)` style merge.
- `tryMatchMarker(text, i)` — walks `MARKERS`, returns `{ elements, nextIndex }` or `null`.
- `appendPlainChar(elements, char)` — the append-or-create plain-element logic.
- `stripEmptyStyles(elements)` — the trailing `map`.

`parseInline()` is now a plain walk: try a marker, else take one plain character, advance. `wordCharBefore`/`wordCharAfter` and their surrogate-pair handling are untouched, and `parseMrkdwn()`'s signature and output are unchanged.

## Why

This is the only writer of the `rich_text` payload for `drafts.create`, so a regression corrupts every draft the CLI creates. At complexity 35 the `_`-only word-boundary rule was inseparable from the walk, and the `matched` boolean coupled the inner loop back to the outer one.

## How it was verified

- The existing 26 tests pass **unchanged** — no test was edited.
- 8 new tests pin the seams the extraction creates: space-padded and one-sided-padded spans, empty markers for every marker, an unterminated `_`, three-level nesting `*_~x~_*`, an outer style over an inner `` `code` `` span, plain text coalescing between two spans, and no plain element between adjacent spans.
- Because this is a pure extraction, reverting the change does **not** make the new tests fail — they pass against the pre-refactor file too. So they were validated by mutation instead: dropping the space-pad guard, the empty-inner guard, `styleSpan`'s recursion, the `wordCharAfter` guard, and `appendPlainChar`'s coalescing each fail the suite, with the new tests among the failures in every case.
- A differential harness against `origin/main` compared `parseMrkdwn()` output over 255,987 inputs — exhaustive over every string up to length 6 from `` ` * _ ~ a ``+space, plus 200,000 random strings up to length 24 including newlines, a combining sequence and an astral letter. Zero mismatches.
- `bun run type-check`, `bun test` (559 pass), `bun run build`, `pre-commit run --all-files` all green.

## Deviations from the issue's HOW

- The HOW named three helpers; two more (`findSpan`, `styleSpan`) were needed to get the guard chain and the recursion out of `tryMatchMarker` — the same pattern as #184.
- `stripEmptyStyles()` is kept, though it is provably unreachable today: no path produces an empty `style` object, so it strips nothing. The HOW names it as a deliverable, so deleting it is a scope change for the maintainer to decide — the misleading comment was replaced with one stating plainly that it is defensive and unreachable. **Deleting it is a reasonable follow-up issue.**
- The HOW suggested testing `tryMatchMarker()` directly. The helpers stay module-private (matching #194) and the new tests drive them through `parseMrkdwn()`, so the refactor does not widen the module's public API. Every acceptance criterion is expressible that way; the mutation runs above show each seam is actually covered.

## Checklist

- [x] Linked to an issue carrying `ready-for-pr`
- [x] Tests added, covering edge cases
- [x] `bun run type-check` and `bun test` pass
- [x] Pre-commit hooks ran on every commit; commits are signed
- [x] No user-facing change → no `CHANGELOG.md` or `docs/` update needed


vibexp-agent:
OK

vibexp-agent:
ok

